PR #3, opened 2020-06-02
AB#851 Refresh tokens after expiry before every API call. If Refresh …

sharpedavid:
[AB#851](https://dev.azure.com/cyrovalente/14004b17-200a-4897-9dd5-d44e111f33a1/_workitems/edit/851) Refresh tokens after expiry before every API call. If Refresh Token expires user will be forced to re-authenticate.

https://dev.azure.com/cyrovalente/IAM%20Project/_workitems/edit/851/

### DIFF
--- a/src/keycloak/index.js
+++ b/src/keycloak/index.js
@@ -11,22 +11,8 @@ let initOptions = {
 keycloak.init(initOptions).then((auth) => {
 
     if (!auth) {
-        throw "Some error";
+        throw "Auth error";
     }
-
-    setInterval(() => {
-        keycloak.updateToken(70).then((refreshed) => {
-            if (refreshed) {
-                console.log('Token refreshed');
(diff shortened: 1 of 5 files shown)


src/keycloak/index.js
```diff
@@ -11,23 +11,9 @@ let initOptions = {
...
-            } else {
-                console.log('Token not refreshed, valid for '
-                    + Math.round(keycloak.tokenParsed.exp + keycloak.timeSkew - new Date().getTime() / 1000) + ' seconds');
-            }
-        }).error(() => {
-            console.log('Failed to refresh token');
-        });
-
-    }, 60000)
-
```
sharpedavid, line 30: I delete this because we don't want to automatically refresh the tokens; we only refresh them when the user is actively clicking button that result in API requests.

PR #5, opened 2020-06-12
Update user attributes

sharpedavid:
I think this completes [task 854](https://dev.azure.com/cyrovalente/IAM%20Project/_sprints/taskboard/IAM%20Project%20Team/IAM%20Project/Sprint%2010?workitem=854).

Changes:

* Update user attributes. 
* Don't render user info until data loads. 
* Fix bug in alert display on user update.

Further comments and explanation in pull request.

### DIFF
--- a/src/components/UserInfo.vue
+++ b/src/components/UserInfo.vue
@@ -1,24 +1,42 @@
+<!--suppress XmlInvalidId -->
 <template>
-  <div id="user-info">
-    <v-alert :value="!!successMessage" type="success" dismissible>{{ successMessage }}</v-alert>
-    <v-alert :value="!!errorMessage" type="error" dismissible>{{ errorMessage }}</v-alert>
+  <div id="user-info" v-if="dataReady">
+    <v-alert v-model="successState" type="success" dismissible>{{ successMessage }}</v-alert>
+    <v-alert v-model="errorState" type="error" dismissible>{{ errorMessage }}</v-alert>
 
     <h1>Update - {{ user.username }}</h1>
-    <v-row no-gutters>
-      <v-col class="col-7">
-        <label for="user-name" class="disabled required">User Name</label>
-        <v-text-field dense outlined disabled id="user-name" v-model="user.username"
-        required :rules="[v => !!v || 'Username is required']"/>
+    <v-card outlined class="subgroup">
+      <h2>User Details</h2>
+      <v-row no-gutters>
+        <v-col class="col-7">
+          <v-form ref="form">
+            <label for="user-name" class="disabled required">User Name</label>
+            <v-text-field dense outlined disabled id="user-name" v-model="user.username"
+                          required :rules="[v => !!v || 'Username is required']"/>
 
-        <label for="first-name" class="required">First Name</label>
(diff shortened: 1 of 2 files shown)


src/components/UserInfo.vue
```diff
@@ -1,24 +1,42 @@
+<!--suppress XmlInvalidId -->
```
sharpedavid, line 1: This is an IntelliJ thing.

src/components/UserInfo.vue
```diff
@@ -1,24 +1,42 @@
...
 <template>
-  <div id="user-info">
-    <v-alert :value="!!successMessage" type="success" dismissible>{{ successMessage }}</v-alert>
-    <v-alert :value="!!errorMessage" type="error" dismissible>{{ errorMessage }}</v-alert>
+  <div id="user-info" v-if="dataReady">
```
sharpedavid, line 3: I use `v-if="dataReady"` to prevent the component from rendering until the data loads. This makes things look a little nicer: it doesn't render empty fields followed by data "popping in".

src/components/UserInfo.vue
```diff
@@ -1,24 +1,42 @@
...
+    <v-alert v-model="successState" type="success" dismissible>{{ successMessage }}</v-alert>
+    <v-alert v-model="errorState" type="error" dismissible>{{ errorMessage }}</v-alert>
```
sharpedavid, line 5: Using `successState` and `errorState` fixes a bug where the messages can only be shown once. That is, if there is an error, and you dismiss the message, subsequent errors won't re-show the message. Same for success: if you update the user once, dismiss the success message, and then update again, the message won't reappear.

src/components/UserInfo.vue
```diff
@@ -1,24 +1,42 @@
...
-        <v-text-field dense outlined id="first-name" v-model="user.firstName" 
-        required :rules="[v => !!v || 'First Name is required']"/>
-
-        <label for="last-name" class="required">Last Name</label>
-        <v-text-field dense outlined id="last-name" v-model="user.lastName"
-        required :rules="[v => !!v || 'Last Name is required']"/>
-      </v-col>
-    </v-row>
+    <v-card outlined class="subgroup">
+      <h2>User Details</h2>
```
sharpedavid, line 9: I put a `v-card` around this just to get the spacing right. Without `v-card` the "Save" button didn't have any room around it. Please go ahead and change this if you have opinions.

BEFORE
image

AFTER
image

src/components/UserInfo.vue
```diff
@@ -76,25 +94,55 @@ export default {
   name: "UserInfo",
   data() {
     return {
+      dataReady: false,
       successMessage: "",
+      successState: false,
       errorMessage: "",
-      user: '',
+      errorState: false,
+      user: { 'attributes': {}},
```
sharpedavid, line 102: If the user doesn't have attributes, Keycloak doesn't return any, which results in some `undefined` exceptions. I give it an empty default value to prevent it.

src/components/UserInfo.vue
tschia: I think it looks better with the card since we have separate save buttons

PR #6, opened 2020-06-15
Set user roles

sharpedavid:


src/api/UsersRepository.js
```diff
@@ -25,6 +25,10 @@ export default {
         return kcRequest().then(axiosInstance => axiosInstance.get(`${resource}/${userId}/${clientRoleMappings}/${clientId}/available`));
     },
 
+    getUserActiveClientRoles(userId, clientId) {
```
sharpedavid, line 28: I'm slow to add documentation, but I think we should add some to help developers understand the difference between all the different role APIs. I think it makes sense to add them in the Repositories because they're like a service classes and will change less often that the UserInfo.vue component.

src/components/UserInfo.vue
```diff
@@ -156,82 +164,84 @@ export default {
...
-
-          this.getUserClientRoles();
-        })
-        .catch(error => {
+        window.scrollTo(0, 0);    
+      }).catch(error => {
           this.errorMessage = this.errorMessage + "Error Updating Roles";
           console.log(error);
-        });
+      });
```
sharpedavid, line 244: UserInfo is only 250 lines long, but I wish could break up the functionality a little. As far as I can see the roles code and the user details code has zero `method` and `data` overlap: that is, they don't use any of the same methods or data. I thought they both used the `user` object, but looking again I don't think they even do that: it uses the `userid` in the path, but never actually GETs or POSTs to that resource.

Anyhow, I wouldn't reject a pull request for this, but if we have time I think we should look into it.

src/components/UserInfo.vue
```diff
@@ -56,31 +56,38 @@
...
+                hide-details="auto"
+                v-for="role in clientRoles"
+                v-model="selectedRoles"
+                :value="role"
+                :label="role.name"
+                v-bind:key="role.name"
+              ></v-checkbox>
+            </v-col>
+            <v-col class="col-4">
+              <label>Effective Roles</label>
```
sharpedavid, line 75: I agree tooltips are a good idea.

src/components/UserInfo.vue
```diff
@@ -156,82 +164,84 @@ export default {
...
+        this.errorState = false;
+        this.successMessage =
             this.successMessage + "Roles Updated Successfully ";
-
-          this.getUserClientRoles();
-        })
-        .catch(error => {
+        window.scrollTo(0, 0);    
+      }).catch(error => {
           this.errorMessage = this.errorMessage + "Error Updating Roles";
```
sharpedavid, line 242: Nitpick or personal preference. I don't know if we have a style guide for this or not, but I feel like the tendency is for modern apps to not use Title Case for messages and alerts.

Gmail:
image

Keycloak:
image

src/components/UserInfo.vue
```diff
@@ -56,31 +56,38 @@
...
+                v-model="effectiveClientRoles"
+                disabled="disabled"
+                :value="role"
+                :label="role.name"
+                v-bind:key="role.name"
+              ></v-checkbox>
+            </v-col>
+          </v-row>
           <div class="my-6" v-if="selectedClientId">
             <v-btn class="secondary" medium v-on:click="updateUserClientRoles()">Save User Role</v-btn>
```
sharpedavid, line 88: Just an idea, but it might be clearer if this were "Save Permissions" or the section was called "Roles" instead of "Permissions".

image

src/components/UserInfo.vue
```diff
@@ -56,31 +56,38 @@
...
-              :label="role.name"
-              v-bind:key="role.name"
-            ></v-checkbox>
-            <v-checkbox
-              hide-details="auto"
-              v-for="role in availableClientRoles"
-              v-model="selectedRoles"
-              :value="role"
-              :label="role.name"
-              v-bind:key="role.name"
```
sharpedavid, line 76: `:` is short for v-bind, and `@` is short for `v-on` I think. Do you have a preference for whether we use the shorthand of not? As a beginner I kind of like reading `v-on` but I'd get use to `@` fast enough.

src/components/UserInfo.vue
```diff
@@ -156,82 +164,84 @@ export default {
...
+        this.successMessage =
             this.successMessage + "Roles Updated Successfully ";
-
-          this.getUserClientRoles();
-        })
-        .catch(error => {
+        window.scrollTo(0, 0);    
+      }).catch(error => {
           this.errorMessage = this.errorMessage + "Error Updating Roles";
           console.log(error);
```
sharpedavid, line 243: I think we should always display a simple error message to the user. I'm familiar with proper client-side error handling and logging, but from a user experience it's nice to see an alert when something goes wrong.

For testing and implementation, I suggest you go to one of the Repositories and add an error:

```javascript
deleteUserClientRoles(userId, clientId, content) 
    let a = true;
    if (a) throw Error('bad stuff')
    //....
}

Then you can test the behaviour of error handling.

src/components/UserInfo.vue
tschia: Agreed, but will do this as a separate piece of work.

sharpedavid:
This is just a nitpick, it wouldn't prevent a merge, but it would be nice if both Roles and Effectives Roles were in the same sort order. It makes lists easier to compare. There's no reason to actually compare these lists though, haha, it's just an instinct, so it's a little distracting.

image

sharpedavid:
I did a very quick smoke test of this and it seems to work. Please go ahead and merge it whenever you're ready.

I do have two questions. First question, why when I sign in with the `testcafe` user do I only see `FMDB` and `IAMDRUPAL` clients? Didn't it used to show the full list of clients, that is, including the default Keycloak clients? For a test, I added a new client `SOMECLIENT`, and it doesn't show up in the Applications list, not even when I assign a role to the user I'm viewing.

Second question, and probably related. Can we use the User Management console to assign another user `user-management`/`realm-management` permissions? Or can that only be done in the Keycloak Admin Console?

PR #9, opened 2020-06-23
Ab#750 event log

sharpedavid:
Add the "Events" and "Admin Events" views. 

On one hand this work is incomplete because we want to add more features, like user ID to username mapping, a more readable display of details (e.g. not JSON), and possibly filtering out events, but on the other hand this code is in a good state to be merged: it's working and has a sensible design to build on. 

If we review and merge this now, future pull requests to add the aforementioned features will be easier to review.

### DIFF
--- a/src/views/AdminEventLog.vue
+++ b/src/views/AdminEventLog.vue
@@ -1,7 +1,10 @@
 <template>
     <div>
-        <button @click="getAdminEvents">Refresh</button>
-
+      <v-text-field
+          v-model="search"
+          append-icon="mdi-magnify"
+          label=" Search"
+      ></v-text-field>
         <v-data-table
                 :headers="headers"
                 :items="adminEvents"
@@ -12,11 +15,13 @@
                 item-key="key"
                 loading-text="Loading events"
                 :loading="loadingStatus"
+                :search="search"
         >
             <template v-slot:expanded-item="{ headers, item }">
                 <td :colspan="headers.length"><pre>{{item.representation | pretty}}</pre></td>
             </template>
         </v-data-table>
+        <button @click="getAdminEvents">Refresh</button>
     </div>
 </template>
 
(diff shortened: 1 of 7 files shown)


src/views/AdminEventLog.vue
```diff
@@ -0,0 +1,81 @@
...
+          getAdminEvents: async function () {
+            this.loadingStatus = true;
+            try {
+              let promise = await AdminEventsRepository.getEvents();
+              this.adminEvents = promise.data;
+              for (let [index, e] of this.adminEvents.entries()) {
+                e.key = index;
+                e.readableDate = formatDate(e.time);
+              }
+            } finally {
```
tschia, line 66: Do we want a catch in case of error?

src/views/EventLog.vue
```diff
@@ -1,5 +1,65 @@
...
+                this.events = promise.data;
+                for (let e of this.events) {
+                  e.readableDate = formatDate(e.time);
+                  if (!e.details) {
+                    e.details = 'no details';
+                  }
+                }
+              } finally {
+                this.loadingStatus = false;
+              }
```
tschia, line 61: Should there be a catch in case of error

src/views/AdminEventLog.vue
```diff
@@ -0,0 +1,81 @@
...
+            <template v-slot:expanded-item="{ headers, item }">
+                <td :colspan="headers.length"><pre>{{item.representation | pretty}}</pre></td>
+            </template>
+        </v-data-table>
+        <button @click="getAdminEvents">Refresh</button>
+    </div>
+</template>
+
+<script>
+    import AdminEventsRepository from "../api/AdminEventsRepository";
```
tschia, line 29: I'm ok to pull out the factory and just import. 
Lets make it consistent across the app.

PR #11, opened 2020-06-26
AB#750 Add event search.

sharpedavid:
This pull request adds search fields to Events and Admin Events. It's pretty simply of course since we're just exposing functionality available via the REST API.

There is still more to do:

* Error handling.
* User ID and client ID correlation.
* Layout and styling. 👈

The layout I just did the bare minimum. We'll figure out what works when we're done.

### DIFF
--- a/src/api/AdminEventsRepository.js
+++ b/src/api/AdminEventsRepository.js
@@ -1,10 +1,17 @@
 import { kcRequest } from "./Repository";
 
-const resource = "/admin-events";
+const resource = "/admin-events" +
+    "?operationTypes=UPDATE" +
+    "&operationTypes=CREATE" +
+    "&operationTypes=DELETE" +
+    "&resourceTypes=USER" +
+    "&resourceTypes=CLIENT_ROLE_MAPPING" +
+    "&max=100";
 
 export default {
 
-    getEvents() {
-        return kcRequest().then(axiosInstance => axiosInstance.get(`${resource}`));
+    getEvents(queryParams) {
+        const resourcePath = queryParams ? `${resource}&${queryParams}` : resource;
+        return kcRequest().then(axiosInstance => axiosInstance.get(resourcePath));
     }
 }
--- a/src/views/AdminEventLog.vue
+++ b/src/views/AdminEventLog.vue
@@ -1,7 +1,22 @@
 <template>
     <div>
       <v-text-field
-          v-model="search"
+          placeholder="User ID"
+          v-model="searchUserId"
+      />
+      <v-text-field
+          placeholder="Date (from)"
+          hint="yyyy-MM-dd"
+          v-model="searchDateFrom"
+      />
+      <v-text-field
+          placeholder="Date (to)"
+          hint="yyyy-MM-dd"
+          v-model="searchDateTo"
+      />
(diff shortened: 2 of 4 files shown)


src/api/AdminEventsRepository.js
```diff
@@ -1,10 +1,17 @@
 import { kcRequest } from "./Repository";
 
-const resource = "/admin-events";
+const resource = "/admin-events" +
+    "?operationTypes=UPDATE" +
+    "&operationTypes=CREATE" +
+    "&operationTypes=DELETE" +
+    "&resourceTypes=USER" +
+    "&resourceTypes=CLIENT_ROLE_MAPPING" +
+    "&max=100";
```
sharpedavid, line 9: During my meeting with Leon we agreed that it makes sense to restrict visible events to "user" events, those events that can be generated by using the User Management Console. (Note that this implies that if a user can do something in the User Management Console that _isn't_ logged, that's possibly a bug.)

src/views/AdminEventLog.vue
```diff
@@ -15,13 +30,13 @@
...
                 :loading="loadingStatus"
-                :search="search"
+                :search="filterEvents"
         >
             <template v-slot:expanded-item="{ headers, item }">
                 <td :colspan="headers.length"><pre>{{item.representation | pretty}}</pre></td>
             </template>
         </v-data-table>
-        <button @click="getAdminEvents">Refresh</button>
+        <button @click="getAllEvents">Refresh</button>
```
sharpedavid, line 39: This "Refresh" button will probably just be removed, actually. Everything can be done with the "Search" button. It was just useful when I first started developing this.

src/views/AdminEventLog.vue
```diff
@@ -50,14 +68,31 @@
...
+            const params = new URLSearchParams();
+            if (this.searchUserId) {
+              params.append('resourcePath', `users/${this.searchUserId}*`);
+            }
+            [
+              {name: 'dateFrom', value: this.searchDateFrom},
+              {name: 'dateTo', value: this.searchDateTo},
+            ].map(param => {
+              if (param.value) params.append(param.name, param.value);
+            });
```
sharpedavid, line 85: This use of the `map` function avoids a long series of `if` statements, which are easy to read but verbose and arguably error prone. Here's it's arguable showy for just two attributes, but makes more sense in the next file, `EventLog.vue`, because there are more fields to map, so I'd like to leave this way it for consistency.

src/views/EventLog.vue
```diff
@@ -25,41 +44,61 @@
...
+        searchEvents: function () {
+          const params = new URLSearchParams();
+          [
+            {name: 'user', value: this.searchUserId},
+            {name: 'client', value: this.searchApplication},
+            {name: 'dateFrom', value: this.searchDateFrom},
+            {name: 'dateTo', value: this.searchDateTo},
+          ].map(param => {
+            if (param.value) params.append(param.name, param.value);
+          });
```
sharpedavid, line 80: This I argue is nice. It's cleaner than four `if` statements, and it's easy to add more search fields.

sharpedavid:
I merged this as agreed upon because Trevor is on vacation, but he'll review all the code when he gets back.

PR #16, opened 2020-07-10
added configurable logout redirect url

tschia:
Confirmed changing the values in `config.json` on the webserver picks up at runtime
We should be able to use this same approach to make `idps_to_show` configurable (still TBD if we just want to use kc_idp_hint instead though)

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -8,14 +8,20 @@ import keycloak from './keycloak';
 import store from './store'
 
 Vue.config.productionTip = false
-
 Vue.prototype.$keycloak = keycloak; //maybe be able to remove this and use the export
 
 keycloak.onAuthSuccess = function () {
-  new Vue({
-    vuetify,
-    router,
-    store,
-    render: h => h(App)
-  }).$mount('#app');
+    fetch(process.env.BASE_URL + "config.json")
+        .then((response) => {
+            return response.json();
+        })
+        .then((config) => {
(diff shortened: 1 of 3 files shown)


src/main.js
```diff
@@ -8,14 +8,20 @@ import keycloak from './keycloak';
...
+        })
+        .then((config) => {
+                Vue.prototype.$config = config
+                new Vue({
+                    vuetify,
+                    router,
+                    store,
+                    render: h => h(App)
+                }).$mount('#app');
+        })
```
sharpedavid, line 26: Just FYI, I did trying using `await` for this and it did work. I'm not so sure about making the `onAuthSuccess` async, but again, it seems to work. If you you want to for some reason, go ahead and make the change (though I didn't test changing the config on the server), otherwise I'd just merge it as is.

```javascript

keycloak.onAuthSuccess = async function () {
    let config = (await fetch(process.env.BASE_URL + "config.json")).json();
    Vue.prototype.$config = config;
    new Vue({
        vuetify,
        router,
        store,
        render: h => h(App)
    }).$mount('#app');
}
```

sharpedavid: Oh, I see Intellij automatically added  `()` around the await. I guess so that I could call `json()` on it. I didn't notice that until just now.

tschia: Thanks, a little cleaner with `await` but it probably warrants a little in depth testing...

PR #17, opened 2020-07-10
AB#1255 Check for missing roles and display error message.

sharpedavid:
https://dev.azure.com/cyrovalente/IAM%20Project/_sprints/taskboard/IAM%20Project%20Team/IAM%20Project/Sprint%2012?workitem=1255

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -44,7 +44,7 @@ export default {
     if (missingRoles.length > 0) {
       this.$store.commit("alert/setAlert", {
         message: `You are missing roles required to use this application.
-          Missing roles: ${missingRoles}.`,
+          Missing roles: ${missingRoles.join(', ')}.`,
         type: "error"
       });
     }
--- a/src/App.vue
+++ b/src/App.vue
@@ -35,8 +35,31 @@ export default {
     return {
       showKeycloakTools: false
     };
+  },
+  mounted() {
+    const missingRoles = getMissingRoles(
+        ["view-events", "manage-users", "view-users", "query-clients", "query-groups", "query-users"],
+        this.$keycloak.tokenParsed.resource_access['realm-management'].roles
+    );
+    if (missingRoles.length > 0) {
+      this.$store.commit("alert/setAlert", {
+        message: `You are missing roles required to use this application.
+          Missing roles: ${missingRoles}.`,
+        type: "error"
+      });
+    }
   }
 };
+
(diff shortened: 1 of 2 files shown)


src/App.vue
```diff
@@ -35,8 +35,31 @@ export default {
...
+        ["view-events", "manage-users", "view-users", "query-clients", "query-groups", "query-users"],
+        this.$keycloak.tokenParsed.resource_access['realm-management'].roles
+    );
+    if (missingRoles.length > 0) {
+      this.$store.commit("alert/setAlert", {
+        message: `You are missing roles required to use this application.
+          Missing roles: ${missingRoles}.`,
+        type: "error"
+      });
+    }
```
sharpedavid, line 50: I think App.vue is a pretty good place for this. It makes it so that no matter which page you visit first (e.g. have bookmarked), the error is shown. The message disappears upon navigation and that's fine too.

src/App.vue
```diff
@@ -35,8 +35,31 @@ export default {
...
+
+function getMissingRoles(rolesRequired, actualRoles) {
+  const missingRoles = [];
+  rolesRequired.forEach(requiredRole => {
+    if (!actualRoles.includes(requiredRole)) {
+      missingRoles.push(requiredRole);
+    }
+  });
+  return missingRoles;
+}
```
sharpedavid, line 62: How to structure Javascript code is new territory for me. I created this function, I think it's "module scoped". I actually wanted to move the code in "mounted" into its own method too, but I don't know where to put it. If you make it a function like this one, you'd need to pass in references to Keycloak and Vuex. If you put it in the Vue Component's "methods" property, I think it will be called reactively by the framework during rendering.

src/keycloak/index.js
```diff
@@ -9,14 +9,6 @@ let initOptions = {
...
-keycloak.init(initOptions).then((auth) => {
-
-    if (!auth) {
-        throw "Auth error";
-    }
-
-}).catch(() => {
-    console.log("Authenticated Failed");
-});
+keycloak.init(initOptions);
```
sharpedavid, line 12: This change isn't required by the PR, but I think it's good. In the event of errors, exceptions are still thrown and logged, I just removed some pretty useless code.

src/App.vue
tschia: discussed on teams. Implementation makes sense to me

PR #19, opened 2020-07-15
Use Vue CLI development mode to show Keycloak tools and all IDPs only when running locally.

sharpedavid:
Use Vue CLI "development" mode to show Keycloak tools and all IDPs only when running locally.

Trevor this works fine for me, but could you please try it out on development/test. The extent of my testing was to add a typo to the string comparison (e.g. `dddevelopment`), but I can't test it on other environments without your help.

This doesn't make the build artifact different, it just uses the NODE_ENV environment variable like is already used in `vue.config.js`.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -7,8 +7,8 @@
         <alert/>
         <router-view></router-view>
       </section>
-      <v-checkbox v-model="showKeycloakTools" :label="`Keycloak Dev Tools`"></v-checkbox>
-      <KeycloakDevTools v-show="showKeycloakTools" />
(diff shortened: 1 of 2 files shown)


src/App.vue
```diff
@@ -33,6 +33,7 @@ export default {
   },
   data() {
     return {
+      vueCliMode: process.env.NODE_ENV,
```
sharpedavid, line 36: I tried referencing NODE_ENV right in the template but it comes back as undefined. Not sure if it's possible to use it outside of `script` tags.

PR #20, opened 2020-07-15
AB#1271 Search User History by Administrator ID.

sharpedavid:
Hi Trevor,

I decided it would be worthwhile to at least replicate what's possible in the Keycloak Admin Console. It's basically free, and it could be useful. You can get the Admin ID out of the "Details" of an Event, or from the User Search screen.

### DIFF
--- a/tests/e2e/alltests.js
+++ b/tests/e2e/alltests.js
@@ -54,7 +54,6 @@ test('Test update user role', async t => {
     await t
         .typeText('#user-search', 'testcafe')
         .click('#search-button')
-        // This is the ID of the testcafe user.
         .click(Selector('td').withText(TEST_CAFE_USER_ID))
         .typeText('#select-client', client, { replace: true })
         .click(Selector('.v-list-item').withText('realm-management'))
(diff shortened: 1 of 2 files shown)


tests/e2e/alltests.js
```diff
@@ -64,5 +63,14 @@ test('Test update user role', async t => {
...
 
+test('Test search by administrator', async t => {
+    await t
+        .click('#admin-event-log-link')
+        .typeText('#admin-id', TEST_CAFE_USER_ID)
+        .pressKey('enter')
+        .expect(Selector('html').textContent)
+        .contains('Test Cafe Cafe');
+});
+
```
sharpedavid, line 74: If we had more tests I'd say organize them in their own files, but this is the only test I've written for the Events view so it's not useful to go in a file by itself. I like what you did for the `createUser.js` tests though.

tests/e2e/alltests.js
```diff
@@ -64,5 +63,14 @@ test('Test update user role', async t => {
         .contains('Roles updated successfully');
 });
 
+test('Test search by administrator', async t => {
+    await t
+        .click('#admin-event-log-link')
+        .typeText('#admin-id', TEST_CAFE_USER_ID)
+        .pressKey('enter')
```
sharpedavid, line 70: Just wanted to point out that I used the `pressKey('enter')` action instead of clicking the search button.

PR #24, opened 2020-09-01
Ab#1206 manage user groups

tschia:
Add ability to set user groups for ITSB Access Team Members
Controls are disabled for any users not in that group (based on a group role mapping for the user-management client)

### DIFF
--- a/src/components/UserUpdateGroups.vue
+++ b/src/components/UserUpdateGroups.vue
@@ -1,0 +1,93 @@
+<template>
+  <v-card outlined class="subgroup">
+    <h2>User Groups</h2>
+    <v-row no-gutters>
+      <v-col class="col-4">
+        <v-checkbox
+            hide-details="auto"
+            v-for="(group, index) in allGroups"
+            :id="'group-' + index"
+            v-model="selectedGroups"
+            :value="group"
+            :label="group.name"
+            :key="group.id"
+        ></v-checkbox>
+      </v-col>
+    </v-row>
+    <div class="my-6">
+      <v-btn id="save-user-groups" class="secondary" medium v-on:click="updateUserGroups()">Save User Groups</v-btn>
+    </div>
+  </v-card>
+</template>
+
+<script>
+import UsersRepository from "@/api/UsersRepository";
+import GroupsRepository from "@/api/GroupsRepository";
+
+export default {
+  name: "UserUpdateGroups",
+  props: ['userId'],
+  data() {
+    return {
+      joinedGroups: [],
+      allGroups: [],
+      selectedGroups: []
+    };
+  },
+  async created() {
+    await this.getUserGroupsData();
+  },
+  methods: {
+    getUserGroupsData: async function() {
+      this.joinedGroups = [];
+      this.allGroups = [];
+      this.selectedGroups = [];
+
+      let userGroupResponses = await Promise.all([UsersRepository.getUserGroups(this.userId), GroupsRepository.get()])
+          .catch(error => {
+            this.$store.commit("alert/setAlert", {
+              message: "Error updating loading user groups: " + error,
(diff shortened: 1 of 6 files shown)


src/components/UserUpdateGroups.vue
```diff
@@ -0,0 +1,106 @@
...
+    };
+  },
+  async created() {
+    this.checkIfAdmin();
+    await this.getUserGroupsData();
+  },
+  methods: {
+    checkIfAdmin: function() {
+      if (this.$keycloak.tokenParsed.resource_access['user-management'] &&
+          this.$keycloak.tokenParsed.resource_access['user-management'].roles.includes('user-management-admin')) {
```
sharpedavid, line 49: Is this a brand new role added just for this feature? Does it needed to be added to the Jenkins client configuration?

src/components/UserUpdateGroups.vue
```diff
@@ -0,0 +1,106 @@
...
+        this.adminUser = true;
+      }
```
sharpedavid, line 51: Not a real issue, but you could add `else { this.adminUser = false }` just to be explicit. Not an actual bug.

src/components/UserUpdateGroups.vue
```diff
@@ -0,0 +1,106 @@
...
+              type: "error"
+            });
+          });
+
+      this.joinedGroups.push(...userGroupResponses[0].data);
+      this.selectedGroups.push(...userGroupResponses[0].data);
+      this.allGroups.push(...userGroupResponses[1].data);
+      // Keycloak "Groups" API returns a different object structure than users/groups
+      // We need them to match for the checkboxes to map
+      this.allGroups = this.allGroups.map(({id,name,path})=>({id,name,path}));
```
sharpedavid, line 71: `map(({id,name,path})->({id,name,path})) looks like it does nothing? Or is it discarding some attributes in the object?

src/components/UserUpdateGroups.vue
```diff
@@ -0,0 +1,106 @@
...
+import UsersRepository from "@/api/UsersRepository";
+import GroupsRepository from "@/api/GroupsRepository";
+
+export default {
+  name: "UserUpdateGroups",
+  props: ['userId'],
+  data() {
+    return {
+      adminUser: false,
+      joinedGroups: [],
```
sharpedavid, line 37: Just a suggestion. Add doc, "Joined groups are the user's current groups". Or maybe just rename it `currentGroups`?

src/components/UserUpdateGroups.vue
tschia: It's discarding a "subgroups" attribute in the object.

src/components/UserUpdateGroups.vue
tschia: yes it is, though I imagine it could be used to limit access to other features within the application at a later time. 
The client config for Jenkins has been updated

src/components/UserUpdateGroups.vue
tschia: Agree, I'll update that. Joined groups is the terminology in the Keycloak admin console, but currentGroups is probably more clear

src/components/UserUpdateGroups.vue
sharpedavid: This is the only comment in the pull request that I really care about.

sharpedavid (approved):

src/components/UserUpdateGroups.vue
sharpedavid: Does that matter?

tschia: yes the vuetify check boxes expects the object structure to match otherwise it won't set any of the check boxes to "checked"

PR #25, opened 2020-09-08
show users linked idps on user details

tschia:
Update to show idp's that are linked to users on the user details page

### DIFF
--- a/src/components/UserDetails.vue
+++ b/src/components/UserDetails.vue
@@ -91,6 +91,14 @@
             />
           </v-form>
         </v-col>
+        <v-col class="col-4" style="margin-left: 30px; padding-left: 20px; border-left: 1px solid #efefef">
+          <label for="linked-idps">Linked Identity Types</label>
+          <ul id="linked-idps" style="margin-top: 5px; list-style: square">
+            <li v-for="identity in user.federatedIdentities" :key="identity.id">
+              {{ formatIdentityProvider(identity.identityProvider) }} [{{ identity.userName }}]
+            </li>
+          </ul>
+        </v-col>
       </v-row>
       <slot></slot>
     </v-card>
@@ -136,6 +144,29 @@ export default {
         .catch(e => {
           console.log(e);
         });
+    },
+    // the IDP name in keycloak doesn't always match what's known by users
+    // format to match standard naming or capitalization
+    formatIdentityProvider: function(idp) {
+      let formattedIdp = idp;
+      switch(idp) {
+        case 'phsa':
+          formattedIdp = 'PHSA';
+          break;
+        case 'moh_idp':
+          formattedIdp = 'MoH LDAP';
+          break;
+        case 'idir':
+          formattedIdp = 'IDIR';
+          break;
+        case 'bceid':
+          formattedIdp = 'BCeID';
+          break;
+        case 'bcsc':
+          formattedIdp = 'BCSC';
+          break;
+      }
+      return formattedIdp;
     }
   },
   computed: {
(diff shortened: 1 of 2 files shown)


src/components/UserDetails.vue
```diff
@@ -136,6 +144,29 @@ export default {
         .catch(e => {
           console.log(e);
         });
+    },
+    // the IDP name in keycloak doesn't always match what's known by users
+    // format to match standard naming or capitalization
+    formatIdentityProvider: function(idp) {
```
tschia, line 150: I think this would be nicer as a computed property but I'm not sure that's possible with an array that will be used in a for loop.

sharpedavid: Something to consider is a "filter". I used on in the AdminEventLog.

https://github.com/bcgov/moh-keycloak-user-management/blob/626bb546741d90d0e67854af9ee6732eef982d3d/src/views/AdminEventLog.vue#L71

Reference: https://vuejs.org/v2/guide/filters.html

src/components/UserDetails.vue
```diff
@@ -136,6 +144,29 @@ export default {
...
+          formattedIdp = 'IDIR';
+          break;
+        case 'bceid':
+          formattedIdp = 'BCeID';
+          break;
+        case 'bcsc':
+          formattedIdp = 'BCSC';
+          break;
+      }
+      return formattedIdp;
```
sharpedavid, line 169: Consider a "lookup table":

```javascript
const formattedIdpTable = {
  'phsa': 'PHSA',
  'moh_idp': 'MoH LDAP',
  'idir': 'IDIR',
  'bceid': 'BCeID',
  'bcsc': 'BCSC'
}

return formattedIdpTable[idp];
```

Reference: https://medium.com/@omwri/javascript-patterns-lookup-tables-26bbaf693e24

src/components/UserDetails.vue
```diff
@@ -263,6 +294,11 @@ export default {
       set(access_team_notes) {
         this.$store.commit("user/setAccessTeamNotes", access_team_notes);
       }
+    },
+    federatedIdentities: {
+      get() {
+        return this.$store.state.user.attributes.federatedIdentities;
```
sharpedavid, line 300: I don't know anything about this, but I'm a little surprised that `federatedIdentities` is under `attributes`. It doesn't seem to be consistent with the other data. Like, in `user.js` you set `federatedIdentities` the same way you set `email`:

```javascript
    setUser(state, user) {
        state.username = user.username
        state.firstName = user.firstName
        state.lastName = user.lastName
        state.email = user.email
        state.enabled = user.enabled
        if (user.attributes) {
            state.attributes.phone = user.attributes.phone
            state.attributes.org_details = user.attributes.org_details
            state.attributes.lockout_reason = user.attributes.lockout_reason
            state.attributes.revoked = user.attributes.revoked
            state.attributes.access_team_notes = user.attributes.access_team_notes
        }
        state.federatedIdentities = user.federatedIdentities;
    },
```

https://github.com/bcgov/moh-keycloak-user-management/blob/626bb546741d90d0e67854af9ee6732eef982d3d/src/store/modules/user.js#L38

tschia: it's not under user attributes? it's outside of the if statement

src/components/UserDetails.vue
tschia: thanks i'll look into this

src/components/UserDetails.vue
sharpedavid: Right, it's not under user attributes in user.js, but in UserDetails.vue it's referenced like this:

```javascript
    federatedIdentities: {
      get() {
        return this.$store.state.user.attributes.federatedIdentities
      }
```

Compare this to how you get username:

```javascript
    username: {
      get() {
        return this.$store.state.user.username;
      },
```

I might be misunderstanding how this works, but it looks odd to me.

PR #27, opened 2020-11-10
Ab#1496 api view clients

ahoplock:
@tschia and I quickly ran through this over Teams already but we want a second set of eyes on it. I can give you a run down later today when you're back online.

**Note:** This does not include integration tests which Trevor will be looking into as part of his api story.

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -1,69 +1,74 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
-	<modelVersion>4.0.0</modelVersion>
-	<parent>
-		<groupId>org.springframework.boot</groupId>
-		<artifactId>spring-boot-starter-parent</artifactId>
-		<version>2.3.4.RELEASE</version>
-		<relativePath/> <!-- lookup parent from repository -->
-	</parent>
-	<groupId>ca.bc.gov.hlth</groupId>
-	<artifactId>moh-ums</artifactId>
-	<version>0.0.1-SNAPSHOT</version>
-	<name>moh-ums</name>
-	<description>Service to call Keycloak User Management API's on behalf of users</description>
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-starter-parent</artifactId>
+        <version>2.3.4.RELEASE</version>
+        <relativePath/> <!-- lookup parent from repository -->
+    </parent>
+    <groupId>ca.bc.gov.hlth</groupId>
+    <artifactId>moh-ums</artifactId>
+    <version>0.0.2-SNAPSHOT</version>
+    <name>moh-ums</name>
+    <description>Service to call Keycloak User Management API's on behalf of users</description>
 
-	<properties>
-		<java.version>11</java.version>
-	</properties>
+    <properties>
+        <java.version>11</java.version>
+    </properties>
 
-	<dependencies>
-		<dependency>
-			<groupId>org.springframework.boot</groupId>
-			<artifactId>spring-boot-starter-webflux</artifactId>
-		</dependency>
-		<dependency>
-			<groupId>org.springframework.boot</groupId>
-			<artifactId>spring-boot-starter-security</artifactId>
-		</dependency>
-		<dependency>
-			<groupId>org.springframework.boot</groupId>
-			<artifactId>spring-boot-starter-oauth2-client</artifactId>
-		</dependency>
-		<dependency>
-			<groupId>org.springframework.security</groupId>
-			<artifactId>spring-security-oauth2-resource-server</artifactId>
-		</dependency>
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-webflux</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-security</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-oauth2-client</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.security</groupId>
+            <artifactId>spring-security-oauth2-resource-server</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.keycloak</groupId>
+            <artifactId>keycloak-admin-client</artifactId>
+            <version>9.0.2</version>
+        </dependency>
 
-		<!-- open api and swagger documentation -->
-		<dependency>
-			<groupId>org.springdoc</groupId>
-			<artifactId>springdoc-openapi-webflux-ui</artifactId>
-			<version>1.4.8</version>
-		</dependency>
+        <!-- open api and swagger documentation -->
+        <dependency>
+            <groupId>org.springdoc</groupId>
+            <artifactId>springdoc-openapi-webflux-ui</artifactId>
+            <version>1.4.8</version>
+        </dependency>
 
 
-		<dependency>
-			<groupId>org.springframework.boot</groupId>
-			<artifactId>spring-boot-starter-test</artifactId>
-			<scope>test</scope>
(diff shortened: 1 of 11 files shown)


backend/pom.xml
```diff
@@ -1,69 +1,74 @@
...
-		</plugins>
-	</build>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.springframework.boot</groupId>
+                <artifactId>spring-boot-maven-plugin</artifactId>
+            </plugin>
+        </plugins>
+    </build>
```
sharpedavid, line 72: Did the pom actually change except for the version number?

ahoplock: Yes, I also added a Keycloak jar to allow for manipulating the response body of ClientRepresentation objects.

```
        <dependency>
            <groupId>org.keycloak</groupId>
            <artifactId>keycloak-admin-client</artifactId>
            <version>9.0.2</version>
        </dependency>
```

I ran formatting in my first commit which explains all the whitespace changes.

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/ClientController.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Base64;
+import java.util.List;
+
+@RestController
+public class ClientController {
+
+    private final WebClientService webClientService;
+    private final Logger logger = LoggerFactory.getLogger(ClientController.class);
+
+    private static final JSONParser jsonParser = new JSONParser(JSONParser.DEFAULT_PERMISSIVE_MODE);
```
sharpedavid, line 26: In SpringBoot, services are singletons by default. It Jakarta EE we used thread-safe pools of EJBs using the `@Stateless` annotation, so we don't have to worry much about thread-safety there. SpringBoot is closer to Apache Camel in that sense, where Processors are singletons.

When dealing with singletons we always have to check the documentation to see if the class field we're adding is thread-safe. For example, a common gotcha is that Java's basic [DateFormat](https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/text/DateFormat.html) class is _not_ thread-safe: "Date formats are not synchronized. It is recommended to create separate format instances for each thread. If multiple threads access a format concurrently, it must be synchronized externally." I remember Greg Perkins finding a bug in Camel route that used DateFormat in a processor, which surfaced only during lots of concurrent requests.

So is JSONParser thread-safe? We don't know. On one hand it would be surprising if it _wasn't_ thread-safe, but on the other hand I find it very surprising that DateFormat isn't thread-safe. So [lacking documentation](http://www.javapractices.com/topic/TopicAction.do?Id=48), it's safest to just assume that the class isn't thread-safe.

The fix is easy and has no cost. Just put the JSONParser right in the method. It might affect performance, but then maybe not.

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/ClientController.java
```diff
@@ -0,0 +1,80 @@
+package ca.bc.gov.hlth.mohums.controller;
+
+import ca.bc.gov.hlth.mohums.webclient.WebClientService;
+import net.minidev.json.JSONArray;
+import net.minidev.json.JSONObject;
+import net.minidev.json.parser.JSONParser;
+import net.minidev.json.parser.ParseException;
```
sharpedavid, line 7: In trying to figure out whether JSONParser was thread-safe, I spent a lot of time trying to figure out where the class was even coming from; like, why is it on the classpath. Here's what I found:

image

So we import spring-boot-starter-oauth2-client, which imports spring-security-oauth2-jose, which imports nimbus-jose-jwt, which imports minidev JSON Parser.

If you Google "Spring Boot parse JSON" you find a lot of recommendations to use Jackson, which is what Spring Boot uses by default to handle JSON return values and method parameters. You get that parser by calling `JsonParserFactory.getJsonParser()`.

I don't recommend you change it at this point, because it works and it's clean, but just wanted to share.

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/ClientController.java
```diff
@@ -0,0 +1,80 @@
...
+    private final Logger logger = LoggerFactory.getLogger(ClientController.class);
+
+    private static final JSONParser jsonParser = new JSONParser(JSONParser.DEFAULT_PERMISSIVE_MODE);
+
+    public ClientController(WebClientService webClientService) {
+        this.webClientService = webClientService;
+    }
+
+    @GetMapping("/clients")
+    public Mono<Object> clients(@RequestHeader("Authorization") String token) {
```
sharpedavid, line 33: This class might be a good place to do some unit testing. It's behaviour is encapsulated and has no tricky dependencies. You pass in a base-64 encoded token, which you can build on the fly, and assert the expected values.

You would need to figure out how to mock `WebClientService`, which I have some ideas for if you like, or you could just factor this bit of code into a new class:

```java
JSONArray roles = parseRoles(token);
List<String> authorizedClients = parseClients(roles);
```

A class that takes a token and returns the authorized clients.

Anyhow, this is just a suggestion. I know we have a lot to learn and a lot to work on,and I don't know what the priority is. Trevor's integration tests will hopefully cover this anyhow, though I would like to also figure out unit testing...

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/ClientController.java
ahoplock: Jackson did cross my mind but I just grabbed what we were using for JSON parsing in one of the POC's.

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/ClientController.java
ahoplock: I can't speak to the thread safety of JSONParser, so I moved it into the method in this commit -> https://github.com/bcgov/moh-keycloak-user-management/pull/27/commits/ef7900cc369a4802503feeb1805a8c79dca3f6ae

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/ClientController.java
```diff
@@ -0,0 +1,79 @@
...
+        JSONArray roles = new JSONArray();
+        String[] parts = token.split("\\.");
+
+        try {
+            JSONObject payload = (JSONObject) jsonParser.parse(decodeBase64(parts[1]));
+            JSONObject resourceAccess = (JSONObject) payload.get("resource_access");
+            JSONObject userManagementService = (JSONObject) resourceAccess.get("user-management-service");
+            roles = (JSONArray) userManagementService.get("roles");
+        } catch (ParseException ex) {
+            logger.error(ex.getMessage());
```
sharpedavid, line 51: Unless it's completely thought through why you want to keep processing the current request, please just `throw new RuntimeException(ex)`. Probably is `parse` is failing something is horribly wrong and we're not going to recover.

backend/src/main/java/ca/bc/gov/hlth/mohums/webclient/WebClientService.java
```diff
@@ -15,6 +19,18 @@ public WebClientService(WebClient kcAuthorizedWebClient) {
         this.kcAuthorizedWebClient = kcAuthorizedWebClient;
     }
 
+    public Mono<Object> getClients(List<String> authorizedClients) {
```
sharpedavid, line 22: This is just my opinion, but my suggestion is to keep `WebServiceClient` "dumb". All it should do is call Keycloak and return the results. If some controller's want to do something with the results, that's they're responsibility. So `ClientController` would be the brains of the operation. That's just my idea, and if you like it as it is that's fine by me.

sharpedavid (approved):

backend/src/main/java/ca/bc/gov/hlth/mohums/webclient/WebClientService.java
ahoplock: Looping in @tschia on this as I followed how he initially set things up for getGroups().

sharpedavid: Sorry I could have been a little clearer. This suggestion was that instead of passing in `authorizedClients` as a method parameter, just have `getClients` return all the clients. Then in the `ClientController` filter the clients there. Then all of the "logic" is happening in one place. Again, this is just a suggestion and a personal preference. The code is fine the way it is.

ahoplock: Gotcha. I moved the filtering logic into the controller in this commit -> https://github.com/bcgov/moh-keycloak-user-management/pull/27/commits/1bb8016d9c05218ccdb1742de2e289115bf0f6bc

tschia: I tend to agree with David, it might be good to try and keep the logic in one place (i.e. the controllers). 
In theory you could then just have a method that's `WebClientService.get(String path)` and have a generic get for all endpoints. (haven't thought out how this might work with request parameters)

PR #29, opened 2020-11-16
Use Keycloak adapter for bearer-only configuration

sharpedavid:
Using Maven dependencies `keycloak-spring-boot-starter` and `keycloak-spring-security-adapter` lets us write slightly less code. It doesn't change the behaviour, but I think it would be our standard set-up for new applications. This application is likely to be used as an example in the future.

I will say there are some pros and cons to this approach, which I can describe if you're interested.


backend/pom.xml
```diff
@@ -19,25 +19,36 @@
...
         </dependency>
+
+        <dependency>
+            <groupId>org.keycloak</groupId>
+            <artifactId>keycloak-spring-boot-starter</artifactId>
+        </dependency>
+
         <dependency>
-            <groupId>org.springframework.security</groupId>
-            <artifactId>spring-security-oauth2-resource-server</artifactId>
```
sharpedavid, line 36: They Keycloak adapter replaces the need to have the `oauth2-resource-server`.

backend/src/main/java/ca/bc/gov/hlth/mohums/security/KeycloakClientRoleConverter.java
```diff
@@ -1,41 +0,0 @@
...
-import org.springframework.security.core.GrantedAuthority;
-import org.springframework.security.core.authority.SimpleGrantedAuthority;
-import org.springframework.security.oauth2.jwt.Jwt;
-
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
-import java.util.stream.Collectors;
-
-public class KeycloakClientRoleConverter implements Converter<Jwt, Collection<GrantedAuthority>> {
```
sharpedavid, line 13: `SimpleAuthorityMapper` in `SecurityConfig` means we can delete this class.

backend/src/main/resources/application.yaml
```diff
@@ -1,6 +1,11 @@
 keycloak:
-  admin-api-url: https://common-logon-dev.hlth.gov.bc.ca/auth/admin/realms/moh_applications
-  base-oauth-url: https://common-logon-dev.hlth.gov.bc.ca/auth/realms/moh_applications
+  realm: moh_applications
+  auth-server-url: https://common-logon-dev.hlth.gov.bc.ca/auth/
+  bearer-only: true
+  use-resource-role-mappings: true
+  resource: user-management
+config:
+  base-oauth-url: https://common-logon-dev.hlth.gov.bc.ca/auth/admin/realms/moh_applications
```
tschia, line 8: Suggest leaving this as admin-api-url, it's not use for authentication it's actually the base url for keycloaks admin API endpoint

sharpedavid: Yes, you're right. I just changed this when I was debugging something and getting a headache.

sharpedavid:
After some discussion, with decided to use native Spring Boot security and OAuth classes.

PR #30, opened 2020-11-18
fixed a defect with our role security implementation

tschia:
Small change to the SecurityConfig that fixes the existing defects related to role based access


ahoplock (approved):
Reviewed and approved (1 minor change I committed to remove an import).

PR #32, opened 2020-11-19
AB#1740 Create user.

sharpedavid:


backend/src/main/java/ca/bc/gov/hlth/mohums/security/SecurityConfig.java
```diff
@@ -35,9 +35,12 @@ public SecurityWebFilterChain securityWebFilterChain(final ServerHttpSecurity ht
                 .pathMatchers("/groups/**").hasRole(viewGroupsRole)
                 .pathMatchers("/users/**").hasRole(viewUsersRole)
                 .pathMatchers("/*").denyAll()
-            .anyExchange().authenticated().and()
-            .oauth2ResourceServer().jwt()
-            .jwtAuthenticationConverter(new ReactiveJwtAuthenticationConverterAdapter(jwtAuthenticationConverter));
+                .anyExchange().authenticated()
+            .and()
+                .oauth2Client()
```
sharpedavid, line 40: I indented the code for clarity, but the new piece is the addition of the `oauth2Client`. In my testing it doesn't make a difference, but all of the examples I've seen include it. Maybe it only matters if you want to customize it, as we have with the `oauth2ResourceServer` and `authorizeExchange`.

backend/src/main/java/ca/bc/gov/hlth/mohums/webclient/WebClientConfig.java
```diff
@@ -26,18 +25,18 @@
...
                         .build();
 
-        DefaultReactiveOAuth2AuthorizedClientManager authorizedClientManager
-                = new DefaultReactiveOAuth2AuthorizedClientManager(clientRegistrationRepository,
-                        authorizedClientRepository);
-
+        var authorizedClientManager =
+                new AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager(
+                        clientRegistrationRepository,
+                        authorizedClientService);
```
sharpedavid, line 39: This change was necessary to prevent a runtime exception when calling `block` on an exchange:

Exception "serverWebExchange cannot be null".

Reference:
* https://github.com/spring-projects/spring-security/issues/8230
* https://github.com/spring-projects/spring-security/issues/7468
* https://gist.github.com/abhi2495/51ad47394da2effc0fb63f97b3a78e37#gistcomment-3390204

backend/src/test/java/ca/bc/gov/hlth/mohums/integration/MoHUmsIntegrationTests.java
```diff
@@ -103,6 +105,19 @@ public void lookupUserAuthorized() throws Exception {
...
+    public void addUser() throws Exception {
+        WebTestClient.ResponseSpec created = webTestClient
+                .post()
+                .uri("/users")
+                .contentType(MediaType.APPLICATION_JSON)
+                .bodyValue("{\"enabled\":true,\"attributes\":{},\"username\":\"bingo\",\"emailVerified\":\"\"}")
+                .header("Authorization", "Bearer " + jwt)
+                .exchange()
+                .expectStatus().isEqualTo(HttpStatus.CONFLICT);
+        // We expect a 409 (Conflict) because the user already exists.
```
sharpedavid, line 118: I know we're use to testing "successful" cases, e.g. where a user is created, but in the case of testing an external API a 409 is good: it shows that the Keycloak API accepted the request, evaluated the body, and determined that the user already exists.

backend/src/main/java/ca/bc/gov/hlth/mohums/security/SecurityConfig.java
```diff
@@ -35,9 +35,12 @@ public SecurityWebFilterChain securityWebFilterChain(final ServerHttpSecurity ht
                 .pathMatchers("/groups/**").hasRole(viewGroupsRole)
                 .pathMatchers("/users/**").hasRole(viewUsersRole)
```
sharpedavid, line 36: I didn't add a `manage-users` role. I'd like to discuss this a bit. We don't need to replicate all of the roles provided by Keycloak. If all admins had the same privileges, we'd only need one role: ADMIN. So what I'm getting at is that we only need roles when we need fine-grained access. Do we need `view-users` and `create-users`, or will `manage-users` suffice?

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
```diff
@@ -51,4 +50,14 @@ public UsersController(WebClientService webClientService) {
...
         return webClientService.get(path, null);
     }
+
+    @PostMapping(value = "/users")
+    public Mono<ResponseEntity<Object>> createUser(@RequestBody Object body) {
+        Mono<ClientResponse> post = webClientService.post("https://common-logon-dev.hlth.gov.bc.ca/auth/admin/realms/moh_applications/users", body);
+        return post.flatMap(response -> Mono.just(
+                ResponseEntity.status(response.statusCode())
+                        .headers(response.headers().asHttpHeaders())
+                        .body(response.bodyToMono(Object.class))));
```
sharpedavid, line 60: Originally I used `block` to get the Keycloak response and write some simple imperative code, but it turns out that you can't call `block` in a reactive context. You'll get a runtime exception, something like "block()/blockFirst()/blockLast() are blocking, which is not supported in thread...". 

It seems that the reactive context uses very few threads and guards them carefully, so it prevents code from deliberately blocking. I didn't notice this at first because you're allowed to block when running the test framework (not a reactive context, I guess), but not when you're running the real application.

Fortunately @tschia pointed me in the right direction with some example code from this blog post: https://blog.jdriven.com/2019/10/spring-security-5-2-oauth-2-exploration-part2/

----

There is one business logic problem here. The current code copies all of the header as-is, but all I need (AFAIK) is the Location header, and that header needs to be modified. We don't want to return `Location: https://common-logon-dev.hlth.gov.bc.ca/auth/admin/realms/moh_applications/users/d862b0ee-1e3f-423b-a200-55f2e8f103d9`, we want to return `Location: https://localhost/users/d862b0ee-1e3f-423b-a200-55f2e8f103d9`

backend/src/main/java/ca/bc/gov/hlth/mohums/security/SecurityConfig.java
ahoplock: HAMIS is likely going to be using `view-users` within the application to get a list of all Keycloak users with a "Data Submitter" role which is currently what they are doing with LDAP. In this case, we would want them to just have view without create.

tschia: Yeah my thinking with the finer grain roles was that we could allow this api to be used outside of just the user management webapp if required.

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
```diff
@@ -51,4 +58,33 @@ public UsersController(WebClientService webClientService) {
         String path = usersPath + "/" + id;
         return webClientService.get(path, null);
     }
+
+    @PostMapping(value = "/users")
```
ahoplock, line 62: Any particular reason you went with `value = ` for @PostMapping here, unlike all the @GetMappings? I'm fine with either, I just like uniformity.

PR #33, opened 2020-11-19
Ab#1496 api view clients unit tests

ahoplock:


backend/pom.xml
```diff
@@ -35,11 +35,6 @@
             <groupId>org.springframework.security</groupId>
             <artifactId>spring-security-oauth2-resource-server</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.keycloak</groupId>
-            <artifactId>keycloak-admin-client</artifactId>
-            <version>9.0.2</version>
-        </dependency>
```
ahoplock, line 42: We can just cast to LinkedHashMap instead of ClientRepresentation which removes the dependency on the Keycloak Admin Client and gets rid of all the null attributes that were being returned due to the casting.

backend/src/test/java/ca/bc/gov/hlth/mohums/util/AuthorizedClientsParserTest.java
```diff
@@ -0,0 +1,43 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class AuthorizedClientsParserTest {
+
+    @Test
+    public void testParse() {
```
ahoplock, line 13: I initially had a test for parsing a live JWT created from code in MoHUmsIntegrationTests but there's no way to assert that the list of **view-client-\*** roles won't change in the future, so I went with just a known JWT.

tschia (approved):

backend/pom.xml
sharpedavid: I'm fine with this, but notably that's a Jackson configuration, whether or not null attributes are serialized. We might want it back one day if we decide to manipulate the response.

backend/pom.xml
```diff
@@ -53,12 +48,6 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-test</artifactId>
             <scope>test</scope>
-            <exclusions>
-                <exclusion>
-                    <groupId>org.junit.vintage</groupId>
-                    <artifactId>junit-vintage-engine</artifactId>
-                </exclusion>
-            </exclusions>
```
sharpedavid, line 61: This is going to be added back, did I read that in the chat? Seeing it removed worries me a little because presumably the Spring Boot team had a good reason for adding it.

ahoplock: When I pasted that back locally to re-test I accidentally put it under the `junit` dependency (an "Adam revert") which worked fine. Now that I reverted it back to being under `spring-boot-starter-test`, I get still see `Tests run: 0, Failures: 0, Errors: 0, Skipped: 0` when executing a `mvn test` through NetBeans.

ahoplock: I re-added it back given our findings in the chat that we're not using JUnit 5 in our test classes. I've also updated the test class to use the new classes.

ahoplock: I've also converted the integration tests to JUnit 5 as discussed in the chat.

PR #37, opened 2020-11-26
refactored webclient service and controllers 

tschia:
specific path information is now in the webclient service

With this refactor the WebClientService class is pretty big. If/when it starts to grow larger should we be looking at creating individual Service classes for each resource type or should we just do this now? (e.g. users, groups, clients)


ahoplock (approved):
Reviewed and approved.

sharpedavid (approved):

PR #38, opened 2020-11-26
AB#1807 Configure User Management Console to use User Management Service

sharpedavid:
[AB#1807](https://dev.azure.com/cyrovalente/14004b17-200a-4897-9dd5-d44e111f33a1/_workitems/edit/1807) Configure User Management Console to use User Management Service

### DIFF
--- a/backend/src/main/resources/application.yaml
+++ b/backend/src/main/resources/application.yaml
@@ -9,6 +9,7 @@ user-management-client:
     manage-users: manage-users
 config:
   vanity-hostname: localhost
+  allowed-origins: "http://local.umc.com:8080"
 spring:
   security:
     oauth2:
@@ -33,3 +34,5 @@ springdoc:
     path: /docs/swagger-ui.html
   api-docs:
     path: /docs/api-docs
+server:
+  port: 9090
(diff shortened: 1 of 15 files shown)


backend/src/main/resources/application.yaml
```diff
@@ -9,6 +9,7 @@ user-management-client:
     manage-users: manage-users
 config:
   vanity-hostname: localhost
+  allowed-origins: "http://local.umc.com:8080"
```
sharpedavid, line 12: These settings need to be modified locally to make things work. I was using `/etc/hosts`/ but I typical set-up would just be `*`.

frontend/env_config/dev/config.json
```diff
@@ -1,4 +1,4 @@
 {
-  "idps_to_show": "idir,moh_idp",
```
sharpedavid, line 2: Unrelated change. We never reference `idps_to_show` so I removed it from all the config files.

frontend/src/views/AdminEventLog.vue
```diff
@@ -74,7 +74,6 @@
             </template>
 
         </v-data-table>
-        <button @click="getAllEvents">Refresh</button>
```
sharpedavid, line 77: Unrelated change. Button used during development, accidentally left in.

ahoplock (approved):
Reviewed and approved. As noted in Teams, I'm not set up to run the UMC locally so I haven't run the code and just reviewed the changes in GitHub.

frontend/env_config/dev/config.json
tschia: We were using it locally for the testcafe user to be able to login I think

PR #45, opened 2020-12-15
Ab#1870 add user client roles

tschia:


backend/src/main/java/ca/bc/gov/hlth/mohums/webclient/WebClientService.java
```diff
@@ -71,6 +71,11 @@ public WebClientService(WebClient kcAuthorizedWebClient) {
         return get(path, null);
     }
 
+    public ResponseEntity<Object> addUserClientRole(String userId, String clientId, Object data) {
+        String path = usersPath + "/" + userId + userClientRoleMappingPath + clientId;
```
sharpedavid, line 75: Not for this PR, but I we should validate the data in query and path parameters to avoid injection style attacks. e.g. If we know a path parameter is supposed to be a valid GUID, check that format before we replay it against Keycloak.

e.g. If we're allowing to POST to `/users/{userId}/role-mappings/clients/{clientGuid}`, maybe that also gives POST access to `/users/{userId}/role-mappings/clients/{clientGuid}/unintended-resource`. (That's just a random bad example. The point is if we're accepting data from clients, we may as well validate it.)

sharpedavid:
Please update and test the UMC? Or is that a different task.

ahoplock (approved):
Reviewed and approved.

sharpedavid (changes requested):
Waiting for UMC changes, but merge whenever you want.

tschia:
UMC Changes added

sharpedavid (approved):

PR #54, opened 2021-01-08
User management console UI improvements

tschia:
Includes:
Warn about SSO logout AB1368
Show role descriptions on hover AB1854
Clean up display of list of roles AB1855

### DIFF
--- a/backend/src/test/java/ca/bc/gov/hlth/mohums/integration/MoHUmsIntegrationTests.java
+++ b/backend/src/test/java/ca/bc/gov/hlth/mohums/integration/MoHUmsIntegrationTests.java
@@ -190,9 +190,9 @@ public class MoHUmsIntegrationTests {
     }
 
     @Test
-    public void addUserGroups() throws Exception {
+    public void removeUserGroups() throws Exception {
         webTestClient
-                .put()
+                .delete()
                 // 123-tschiavo user
                 .uri("users/39f73cbd-dbf0-41c6-a45c-997c44c1c952/groups/1798203d-027f-4856-a445-8a90c1dc9756")
                 .header("Authorization", "Bearer " + jwt)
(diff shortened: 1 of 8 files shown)


backend/src/test/java/ca/bc/gov/hlth/mohums/integration/MoHUmsIntegrationTests.java
```diff
@@ -200,6 +200,17 @@ public void addUserGroups() throws Exception {
...
+    @Test
+    public void removeUserGroups() throws Exception {
+        webTestClient
+                .delete()
+                // 123-tschiavo user
+                .uri("users/39f73cbd-dbf0-41c6-a45c-997c44c1c952/groups/1798203d-027f-4856-a445-8a90c1dc9756")
+                .header("Authorization", "Bearer " + jwt)
+                .exchange()
+                .expectStatus().isEqualTo(HttpStatus.NO_CONTENT); //204 indicates success
+    }
```
sharpedavid, line 212: If we're deleting this user, how will it be recreated for other tests? Or are we not bothering.

frontend/src/components/UserUpdateRoles.vue
```diff
@@ -105,6 +140,7 @@ export default {
         this.getUserAvailableClientRoles(),
         this.getUserActiveClientRoles()
       ]);
+      // TODO roles that are in effective but not active should not be included in clientRoles
```
sharpedavid, line 143: Are there are consequences to this? Is it inconvenient for the users? Should we "fix" it before going to prod?

sharpedavid (approved):

backend/src/test/java/ca/bc/gov/hlth/mohums/integration/MoHUmsIntegrationTests.java
tschia: we aren't deleting the user just the group association, the comment could be made better by also including the group name. 
It still gives a 204 regardless of if the group is there to remove or not

frontend/src/components/UserUpdateRoles.vue
tschia: There's some weird behavior if a role is added to a user directly, and then also provisioned by group or composite. Its shows the role as available to remove/assign but you cant actually do anything with it,

This is kind of a product of how the keycloak apis work (the admin console has the same behavior). It shouldn't really impact anyone with our production use cases.

PR #56, opened 2021-01-29
Re: #2196 - Implement search by Organization in the user management console

gusslima:
Task #2247 - API Development.

The following changes introduce a "filter by Org. ID" feature when a request to search Users is submitted to UMS backend.

The new feature introduces two new classes:
1. Generic class used to filter a response: https://github.com/gusslima/moh-keycloak-user-management/blob/master/backend/src/main/java/ca/bc/gov/hlth/mohums/util/FilteredResponseEntities.java
2. Concrete logic to filter Users by Org. ID: https://github.com/gusslima/moh-keycloak-user-management/blob/master/backend/src/main/java/ca/bc/gov/hlth/mohums/util/FilterUserByOrgId.java

The following are new integration tests in https://github.com/gusslima/moh-keycloak-user-management/blob/master/backend/src/test/java/ca/bc/gov/hlth/mohums/integration/MoHUmsIntegrationTests.java (lines 114 thru 199)
* searchByOrganization()
* searchByEmailAndOrganization()
* searchByNonExistingOrganization() <-- this test assumes that the HTTP status returned when no users are found is NOT_FOUND.


sharpedavid:
Hi Gus, after a quick scan this looks good. 

I'm going to take some time to nitpick through this. Maybe I'll find something actually important while I do, but a lot of the feedback is just to make the code consistent with the style we use on MoH projects. The style isn't written in stone, so feel free to give feedback if you think it's important.

If I find a bug or requirements issue, I'll say so, but otherwise everything else is a request for minor revisions.

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
```diff
@@ -48,7 +53,19 @@ public UsersController(WebClientService webClientService, @Value("${config.vanit
         search.ifPresent(searchValue -> queryParams.add("search", searchValue));
         username.ifPresent(usernameValue -> queryParams.add("username", usernameValue));
 
-        return webClientService.getUsers(queryParams);
+        final ResponseEntity<Object> users = webClientService.getUsers(queryParams);
+        final ResponseEntity<Object> filteredUsers;
+
+        if (org.isPresent()) {
+            final Predicate<Object> byOrganizationId = org.map(FilterUserByOrgId::new).get();
```
sharpedavid, line 60: This would be simpler as `new FilterUserByOrgId(org.get());`.

gusslima:
Hi David,

Thanks for the review. Is there any place where the coding style of MoH projects is described or outlined? Confluence, Ensemble, GitHub, or a BC Gov site??

Thanks,

Gustavo Lima | Senior Consultant
CGI
T: 780-401-2253

sharpedavid:
Hi Gus,

Technically yes, there is a style guide, but practically no, it hasn't been
read or updated in many years. I don't know where it is, and I wouldn't
refer to it if I did!

The most important thing I'd say, especially as the code base grows, is to
make sure the new code is consistent with the existing code. If you have
questions about what style has been used on other MoH projects, Trevor,
Adam, and I have worked on several MoH projects over the years. I will say
there is a great deal of variability between projects though. The style is
usually established by the first few devs on the project and carried on
from there.

Regards,
David

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
```diff
@@ -48,7 +53,19 @@ public UsersController(WebClientService webClientService, @Value("${config.vanit
         search.ifPresent(searchValue -> queryParams.add("search", searchValue));
         username.ifPresent(usernameValue -> queryParams.add("username", usernameValue));
 
-        return webClientService.getUsers(queryParams);
+        final ResponseEntity<Object> users = webClientService.getUsers(queryParams);
+        final ResponseEntity<Object> filteredUsers;
```
sharpedavid, line 57: I know opinions on this vary, but we're aren't using the `final` keyword in the project _except for class fields_. It ends up cluttering up the code, as most variables end up being `final`.

backend/src/main/java/ca/bc/gov/hlth/mohums/util/FilterUserByOrgId.java
```diff
@@ -0,0 +1,89 @@
...
+    private final String orgId;
+    private final JSONParser jsonParser;
+    private final Expression expression;
+
+    public FilterUserByOrgId(final String id) {
+        this.orgId = id;
+        this.jsonParser = new JSONParser(JSONParser.DEFAULT_PERMISSIVE_MODE);
+        this.expression = new SpelExpressionParser()
+                .parseExpression("#this['attributes']['org_details']");
+    }
```
sharpedavid, line 38: I would move these assignments to the field declarations. There is only one constructor and these never change.

backend/src/main/java/ca/bc/gov/hlth/mohums/util/FilterUserByOrgId.java
```diff
@@ -0,0 +1,89 @@
...
+    
+    private static final Logger LOGGER = LoggerFactory.getLogger(AuthorizedClientsParser.class);
+
+    private final String orgId;
+    private final JSONParser jsonParser;
+    private final Expression expression;
+
+    public FilterUserByOrgId(final String id) {
+        this.orgId = id;
+        this.jsonParser = new JSONParser(JSONParser.DEFAULT_PERMISSIVE_MODE);
```
sharpedavid, line 35: I don't think JSONParser is thread-safe. The class doesn't say whether it is, which usually means we assume it's not. Also, I found this in the author's GitHub:

https://github.com/netplex/json-smart-v2/blob/master/json-smart/src/main/java/net/minidev/json/parser/JSONParserString.java#L23

To avoid issues, make it a local variable in `extractOrgId`. (Then that method can also become `private static`.)

backend/src/main/java/ca/bc/gov/hlth/mohums/util/FilterUserByOrgId.java
```diff
@@ -0,0 +1,89 @@
...
+        this.expression = new SpelExpressionParser()
+                .parseExpression("#this['attributes']['org_details']");
+    }
+    
+    Logger getLogger() {
```
sharpedavid, line 40: In other classes we just reference the `logger` field directly.

backend/src/main/java/ca/bc/gov/hlth/mohums/util/FilteredResponseEntities.java
```diff
@@ -0,0 +1,89 @@
...
+        return this;
+    }
+
+    public ResponseEntity<Object> toResponseEntity() {
+        final ResponseEntity<Object> responseEntity;
+        
+        if (sourceStatusCode == OK_STATUS_CODE) {
+            if (CollectionUtils.isEmpty(entities)) {
+                // REVIEW - Is the frontend expecting a 404 (NOT_FOUND) or a 204 (NO_CONTENT)
+                // HTTP status code when no results are found?
```
sharpedavid, line 75: Our goal is to imitate the behaviour of the Keycloak REST API as much as possible, so for these questions I use the Keycloak Admin Console and see what the result is. In this case I learned that Keycloak will return a 200 with an empty JSON list.

image

backend/src/main/java/ca/bc/gov/hlth/mohums/util/FilteredResponseEntities.java
```diff
@@ -0,0 +1,89 @@
...
+        else if (responseBody instanceof Collection) {
+            final Collection<?> collection = (Collection) responseBody;
+            entities = collection.stream().filter(Objects::nonNull).collect(Collectors.toList());
+        }
+        else {
+            entities = Collections.singleton(responseBody);
+        }
+        
+        return entities;
+    }
```
sharpedavid, line 61: I suspect this class can be removed or simplified. I'll try to explain in a comment, but I can also screen-share on Monday.

Basically instead of `extractEntities` you can let Spring do the conversion for you. In `WebClientService` we have this method:

```java
private ResponseEntity<Object> get(String path, MultiValueMap<String, String> queryParams) {
    return kcAuthorizedWebClient
            .get()
            .uri(t -> t
                    .path(path)
                    .queryParams(queryParams)
                    .build())
            .exchange().block().toEntity(Object.class).block();
}
```

This works for most of our use cases, but it's inconvenient here where we really want a `List`. Spring can handle that though! Add a new method:

```java
private ResponseEntity<List<Object>> getList(String path, MultiValueMap<String, String> queryParams) {
    return kcAuthorizedWebClient
            .get()
            .uri(t -> t
                    .path(path)
                    .queryParams(queryParams)
                    .build())
            // Here we use toEntityList
            .exchange().block().toEntityList(Object.class).block();
}
```

Use this `getList` method in `getUsers` instead of `get`. Then in `UsersController` you can call:

```java
if (org.isPresent()) {
    List<Object> filteredUsers = users.getBody().stream().filter(new FilterUserByOrgId(org.get())).collect(Collectors.toList());
    return ResponseEntity.status(users.getStatusCode()).body(filteredUsers);
}
```

Because we don't need to change the status code for an empty list, I don't think there is much need for extra handling and the entire `FilteredResponseEntities` can be deleted.

I tested this locally and it seems to work, and the test cases still pass.

backend/src/test/java/ca/bc/gov/hlth/mohums/integration/MoHUmsIntegrationTests.java
```diff
@@ -91,12 +94,107 @@ public void clientsAuthorized() throws Exception {
...
+                        .queryParam("org", "00001763")
+                        .build())
+                .header("Authorization", "Bearer " + jwt)
+                .exchange()
+                .expectStatus().isOk()
+                .expectBodyList(Object.class)
+                .returnResult()
+                .getResponseBody();
+
+        // ... should return at least one filtered user
```
sharpedavid, line 180: Nice test and good use of commenting. I try not to "over-comment", but these really add clarity to your test case.

backend/src/test/java/ca/bc/gov/hlth/mohums/integration/MoHUmsIntegrationTests.java
```diff
@@ -304,4 +402,9 @@ public void clientsNoJwtUnauthorized() throws Exception {
                 .exchange()
                 .expectStatus().isUnauthorized(); //HTTP 401
     }
+
+    WebTestClient getClientForDebug() {
```
sharpedavid, line 406: Please delete this if we don't need it.

backend/src/main/java/ca/bc/gov/hlth/mohums/util/FilterUserByOrgId.java
```diff
@@ -0,0 +1,89 @@
...
+import org.springframework.expression.spel.support.StandardEvaluationContext;
+
+/**
+ *
+ */
+public final class FilterUserByOrgId implements Predicate<Object> {
+
+    private static final int ABBREVIATED_USER_ENTITY_LENGTH = 64;
+    
+    private static final Logger LOGGER = LoggerFactory.getLogger(AuthorizedClientsParser.class);
```
sharpedavid, line 27: Logger name should not be `AuthorizedClientsParser`.

backend/src/main/java/ca/bc/gov/hlth/mohums/util/FilterUserByOrgId.java
```diff
@@ -0,0 +1,89 @@
...
+    
+    @Override
+    public boolean test(final Object userEntity) {
+        final EvaluationContext context = new StandardEvaluationContext(userEntity);
+
+        Collection<?> orgDetails = Collections.emptyList();
+
+        try {
+            orgDetails = expression.getValue(context, Collection.class);
+        } catch (final ExpressionException e) {
```
sharpedavid, line 52: Unless we know exactly why the exception is thrown and want to handle that case explicitly, we don't catch exceptions, we just re-throw them as unchecked exceptions, so `throw new RuntimeException(e)`.

backend/src/main/java/ca/bc/gov/hlth/mohums/util/FilterUserByOrgId.java
```diff
@@ -0,0 +1,89 @@
...
+    String extractOrgId(final Object orgDetails) {
+        String id = StringUtils.EMPTY; // Default value.
+
+        final String organizationDetails = Objects.toString(orgDetails, "{}");
+
+        try {
+            final JSONObject organization = (JSONObject) jsonParser.parse(organizationDetails);
+            id = organization.getOrDefault("id", id).toString();
+        } catch (final ParseException | RuntimeException e) {
+            getLogger().warn("Unable to parse: \"{}\": {}.", organizationDetails, e.toString());
```
sharpedavid, line 74: In the logs I see this message:
```
Unable to parse: "123456": java.lang.ClassCastException: class java.lang.Integer cannot be cast to class net.minidev.json.JSONObject (java.lang.Integer is in module java.base of loader 'bootstrap'; net.minidev.json.JSONObject is in unnamed module of loader 'app').
```

I tracked this down to user [blahblah](https://common-logon-dev.hlth.gov.bc.ca/auth/admin/master/console/#/realms/moh_applications/users/17c6ad38-0f56-4372-b880-e7124b43dd7b/user-attributes) which has `org_details=123456`: an Integer, not JSON.

So in this case I guess it makes sense to catch the exception because we can't stop people from creating `org_details` attributes that don't contain JSON.

sharpedavid (changes requested):
Review complete. Please see my comments.

backend/src/main/java/ca/bc/gov/hlth/mohums/util/FilterUserByOrgId.java
```diff
@@ -0,0 +1,83 @@
...
+    public boolean test(Object userEntity) {
+        EvaluationContext context = new StandardEvaluationContext(userEntity);
+        Collection<?> orgDetails = Collections.emptyList();
+
+        try {
+            orgDetails = EXPRESSION.getValue(context, Collection.class);
+        } catch (SpelEvaluationException e) {
+            // This exception is thrown when the evaluated user entity
+            // does not contain the nested entries identified by EXPRESSION.
+            if (LOGGER.isDebugEnabled()) {
```
sharpedavid, line 50: Nitpick.

The convention of checking `isDebugEnabled` before logging to DEBUG is a potential performance optimization, but we try to avoid "premature optimization" on our projects. In this case, the API we're developing will not see heavy use. We have a small, fixed set of administrators manually administering user permissions. If we have performance issues one day, we'll measure and optimize where needed. (My guess is that, if that comes, Keycloak and the database will be the problem, not this simple API proxy.)

Also, in this case it's unlikely to improve performance. We're using String parameterization in the log message, not String concatenation, and I don't believe it's actually executed unless DEBUG logging is enabled. https://stackoverflow.com/a/653846/201891

backend/src/main/java/ca/bc/gov/hlth/mohums/util/FilterUserByOrgId.java
```diff
@@ -0,0 +1,83 @@
...
+                LOGGER.debug("Unable to filter \"{}\" by Org. ID {}: {}.",
+                        abbreviateEntity(userEntity), orgId, e.toString());
+            }
+        }
+
+        return !CollectionUtils.isEmpty(orgDetails) && orgDetails
+                .stream()
+                .map(FilterUserByOrgId::extractOrgId)
+                .anyMatch(id -> StringUtils.equalsIgnoreCase(orgId, id));
```
sharpedavid, line 59: Nitpick.

One of the advantages of working with streams (or iterators) is that you don't need to check for "empty". The Javadoc for `anyMatch` reads "If the stream is empty then false is returned and the predicate is not evaluated."

gusslima:
Can getValue() return null?

Gustavo Lima | Senior Consultant
CGI
T: 780-401-2253

backend/src/main/java/ca/bc/gov/hlth/mohums/util/FilterUserByOrgId.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+    
+    static String extractOrgId(Object orgDetails) {
+        String organizationDetails = Objects.toString(orgDetails, "{}");
+        JSONParser jsonParser = new JSONParser(JSONParser.DEFAULT_PERMISSIVE_MODE);
+        String id = StringUtils.EMPTY; // Default value.
+
+        try {
+            id = ((JSONObject) jsonParser.parse(organizationDetails)).getOrDefault("id", id).toString();
+        } catch (ParseException | RuntimeException e) {
```
sharpedavid, line 69:
```suggestion
        } catch (ParseException e) {
```

Originally I intended to that you should _rethrow_ the ParseException as a RuntimeException, not catch RuntimeException, but on second thought, what you had is fine. Just catch the ParseException.

backend/src/main/java/ca/bc/gov/hlth/mohums/util/FilterUserByOrgId.java
```diff
@@ -0,0 +1,83 @@
...
+            // These exceptions are thrown when the value found by EXPRESSION
+            // is not a valid JSON object. We can't stop people from creating
+            // 'org_details' attributes that don't contain JSON data.
+            LOGGER.warn("Unable to parse: \"{}\": {}.", organizationDetails, e.toString());
```
sharpedavid, line 73:
```suggestion
            LOGGER.warn("Unable to parse: \"{}\": {}.", organizationDetails, e);
```

Log the original stack trace, not just the results of `toString`.

sharpedavid (changes requested):
I added a few "nitpick" comments, but the important thing now I think is just to add a _small_ amount of Javadoc to `FilterUserByOrgId`. Because we're just working with `Object` and JSON, the code is a little hard to follow. I was thinking something like this:

> `FilterUserByOrgId` returns `true` if the organization ID provided in the constructor matches the organization ID found in the given Object. The given Object must be a Java Map containing keys "attributes" > "org_details", and "org_details" must be JSON containing something like `{"id":"00001763","name":"Interior Health Authority"}`. If the ID given in the constructor matches the ID in the JSON, the `FilterUserByOrgId` returns `true`, otherwise it returns `false`. If the Object or JSON structure structure is unexpected, it returns `false`.

sharpedavid (approved):

PR #57, opened 2021-02-03
Ab#2196 advanced search organization

tschia:
initial implementation of advanced search and search by org

### DIFF
--- a/frontend/src/components/UserSearch.vue
+++ b/frontend/src/components/UserSearch.vue
@@ -27,7 +27,7 @@
         />
       </v-col>
       <v-col class="col-4">
-        <v-btn id="search-button" class="secondary" medium @click.native="searchUser(userSearchInput)">Search Users</v-btn>
+          <v-btn id="search-button" class="secondary" medium @click.native="searchUser('&search='+userSearchInput)">Search Users</v-btn>
       </v-col>
       <v-col class="col-2">
         <v-btn id="create-user-button" class="success" medium @click.native="goToCreateUser">Create New User</v-btn>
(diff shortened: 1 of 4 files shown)


tschia:
Still need to update the UI tests - just looking for some initial feedback

frontend/src/components/UserSearch.vue
```diff
@@ -92,6 +191,35 @@ export default {
...
+    searchUserAdvanced: function() {
+      let params = '';
+      params = this.addQueryParameter(params, "lastName", this.lastNameInput)
+      params = this.addQueryParameter(params, "firstName", this.firstNameInput)
+      params = this.addQueryParameter(params, "username", this.usernameInput)
+      params = this.addQueryParameter(params, "email", this.emailInput)
+      params = this.addQueryParameter(params, "org", this.organizationInput)
+
+      UsersRepository.get(
+          "?briefRepresentation=true&first=0&max=300" + params
```
gusslima: Using briefRepresentation=true makes the search by Org. ID yield no results since the UserRepresentation returned by KC does not contain 'attributes'. Should this be overridden by the API when "org" parameter is present?

gusslima: never mind. I see that the final commit uses briefRepresentation=false

tschia: No sorry, that's a mistake by me. I messed up my rebase and lost some of my changes, I thought I fixed it all but I missed this. I'll update to set briefRepresentation=false

sharpedavid (approved):

frontend/tests/e2e/tests/advancedSearch.js
```diff
@@ -0,0 +1,56 @@
+import { itsbAccessTeamUser } from "../roles/roles"
+import AdvancedSearchForm from "../pages/AdvancedSearchForm";
+
+const SITE_UNDER_TEST = 'http://localhost:8080/#/users/';
```
sharpedavid, line 4: It would be nice to automatically discover the port, but that's fine for now.

PR #58, opened 2021-02-08
Ab#2197 advanced search user role

gusslima:
Preliminary (inefficient) solution for Search By Role, combined with other criteria specified in the Advanced Search form.

### DIFF
--- a/backend/src/test/java/ca/bc/gov/hlth/mohums/integration/MoHUmsIntegrationTests.java
+++ b/backend/src/test/java/ca/bc/gov/hlth/mohums/integration/MoHUmsIntegrationTests.java
@@ -1,8 +1,16 @@
 package ca.bc.gov.hlth.mohums.integration;
 
+import java.io.IOException;
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.util.List;
+import java.util.Map;
 import net.minidev.json.JSONObject;
 import net.minidev.json.parser.JSONParser;
 import net.minidev.json.parser.ParseException;
+import org.assertj.core.api.Assertions;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
@@ -17,14 +25,6 @@ import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.test.context.junit.jupiter.SpringExtension;
 import org.springframework.test.web.reactive.server.WebTestClient;
-
-import java.io.IOException;
-import java.net.URI;
-import java.net.http.HttpClient;
-import java.net.http.HttpRequest;
-import java.net.http.HttpResponse;
-import java.util.List;
-import org.assertj.core.api.Assertions;
 
 @ExtendWith(SpringExtension.class)
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
@@ -70,7 +70,7 @@ public class MoHUmsIntegrationTests {
 
         return access_token;
     }
-    
+
     @Test
     public void groupsAuthorized() throws Exception {
         webTestClient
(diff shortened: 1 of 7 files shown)


backend/src/test/java/ca/bc/gov/hlth/mohums/integration/MoHUmsIntegrationTests.java
```diff
@@ -262,15 +308,15 @@ public void deleteUserClientRole() {
...
-                        "]")
+                .bodyValue("[\n"
+                        + "    {\n"
+                        + "        \"id\": \"a88f491a-3bd1-46ce-9cf6-c509f9a916f8\",\n"
+                        + "        \"name\": \"PSDADMIN\",\n"
+                        + "        \"composite\": false,\n"
+                        + "        \"clientRole\": true,\n"
+                        + "        \"containerId\": \"db9dd8ab-0f38-4471-b396-e2ddac45a001\"\n"
+                        + "    }\n"
+                        + "]")
```
sharpedavid, line 319: Please avoid including formatting changes in a PR. It makes it harder to figure out what the actual changes are.

(We do need to standardize on our formatting. We have some people in Intellij and some in Netbeans. In the past we've used a Maven-based formatter, but haven't set it up for this project.)

frontend/src/api/ClientsRepository.js
```diff
@@ -10,6 +10,11 @@ export default {
     getRoles(clientId){
         return kcRequest().then(axiosInstance => axiosInstance.get(`${resource}/${clientId}/roles`));
     },
+    
+    // REVIEW - Do we need this API in UMS, or should it use the KC API?
```
sharpedavid, line 14: We actually want to entirely remove the UMC's dependency on the KC API. Our reasons for doing so are documented in [backend/README.md](https://github.com/bcgov/moh-keycloak-user-management/blob/master/backend/README.md).

frontend/src/api/ClientsRepository.js
```diff
@@ -10,6 +10,11 @@ export default {
     getRoles(clientId){
         return kcRequest().then(axiosInstance => axiosInstance.get(`${resource}/${clientId}/roles`));
```
sharpedavid, line 11: Does this API exist in the UMS now? If so, we should use it instead of the Keycloak API. Then ClientsRepository no longer depends on the KC API.

frontend/src/components/UserSearch.vue
```diff
@@ -96,16 +96,19 @@
...
       <v-col class="col-6">
         <label for="adv-search-role">
           Role
         </label>
-        <v-text-field
+        <v-autocomplete
             id="adv-search-role"
+            v-model="roleInput"
+            :items="clientRoles"
+            item-value="id"
```
sharpedavid, line 111: It's very difficult to clear the value after choosing something. If you delete the field contents and click/tab out of the field, the contents reappear. The only thing I found that works it to delete the field contents and then press "Esc".

sharpedavid: 🐛

There also seems to be a bug here. If I enter a role, e.g. "GIS : ADMIN", and run a search, I get results. But then if I delete the field contents (delete "GIS : ADMIN" and press Esc), and then re-reun the search, no search is executed: no request is sent, but the datatable shows the loading indicator. If I then enter a role, I can search successfully again.

frontend/src/components/UserSearch.vue
```diff
@@ -138,8 +141,31 @@
 
 <script>
 import UsersRepository from "@/api/UsersRepository";
+import ClientsRepository from "@/api/ClientsRepository";
 import organizations from "@/assets/organizations"
 
+// REVIEW - Is this the best approach to define custom objects using Vue?
+//   Should the custom object be defined in a more Vue-compliant way?
```
sharpedavid, line 148: I think you're the first to define a class in the project that isn't just a helper class for Testcafe.

I don't have any experience with this, so I'm not sure what to say. A question though, was this done just so that we could have the `toString` method, which is just to show the items in the drop-down list?

frontend/src/components/UserSearch.vue
```diff
@@ -138,8 +141,31 @@
...
+    this.id = role.id;
+    this.name = role.name;
+    this.description = role.description;
+    this.clientId = role.containerId;
+    this.clientName = clients.find(client => this.clientId === client.id).name;
+  }
+
+  toString() {
+    let s = `${this.clientName} : ${this.name}`;
+    if (typeof (this.description) !== 'undefined') {
```
sharpedavid, line 161: I'm not a Javascript expert, but I think you can just use:

```
if (this.description) {
```

It covers null, undefined, and the empty String.

frontend/src/components/UserSearch.vue
```diff
@@ -153,18 +179,26 @@ export default {
...
       organizationInput: "",
+      roleInput: "",
       searchResults: [],
       userSearchLoadingStatus: false,
       advancedSearchSelected: false
     };
   },
+  async created() {
+    // TODO error handling
+    // REVIEW - Should this be pushed to the "$store" once it is loaded?
```
sharpedavid, line 199: Let's see how it performs on DEV and TEST first.

frontend/src/components/UserSearch.vue
```diff
@@ -208,6 +255,45 @@ export default {
...
       return parameters;
+    },
+    getClientsAndRoles: function() {
+      return ClientsRepository.get()
+        .then(response => {
+          this.clients = response.data;
+          this.getClientRoles();
+        })
+        .catch(e => {
+          console.log(e);
```
sharpedavid, line 266: We need some error handling here I think? Display the error the same as the others?

frontend/src/components/UserSearch.vue
```diff
@@ -208,6 +255,45 @@ export default {
...
+      Promise.all(getRolesRequests)
+        .then(roles => this.setClientRoles(roles));
+    },
+    setClientRoles: function(roles) {
+      let allRoles = roles
+        .flatMap(roles => roles.data)
+        .map(role => new ClientRole(role, this.clients));
+      allRoles.sort((a,b) => a.toString().localeCompare(b.toString()));
+      this.clientRoles.push(...allRoles);
+    },
```
sharpedavid, line 281: This is a suggestion, but I find it actually more confusing to have the logic spread out between `getClientsAndRoles`, `getClientRoles`, and `setClientRoles`. When you're using promise chaining and functional methods, I think it's easier to read it one continuous line. Basically replace all three of those methods with this one method:

```javascript
loadClientsAndRoles: function () {
  return ClientsRepository.get()
      .then(response => {
        this.clients = response.data;
        let roleRequests = this.clients.map(client => ClientsRepository.getRoles(client.id));
        Promise.all(roleRequests)
            .then(roles => {
              let allRoles = roles
                  .flatMap(roles => roles.data)
                  .map(role => new ClientRole(role, this.clients))
                  .sort((a, b) => a.toString().localeCompare(b.toString()));
              this.clientRoles.push(...allRoles);
            });
      })
      .catch(e => {
        console.log(e);
      });
},
```
If we were to keep these three methods, I would rename them. Naming them `get` and `set` brings to mind accessors, but they're not really accessors, they're initialization:

* `getClientsAndRoles` -> `loadClientsAndRoles`
* `getClientRoles` -> `loadClientRoles`
* `setClientRoles` -> `addRoles` maybe?

frontend/src/components/UserSearch.vue
```diff
@@ -187,12 +221,25 @@ export default {
...
+      let roleId = this.roleInput.trim();
 
       UsersRepository.get(
         "?briefRepresentation=false&first=0&max=300" + queryParameters
       )
         .then(response => {
-          this.searchResults = response.data;
+          let results = response.data;
+          if (this.advancedSearchSelected && roleId !== "") {
+            // REVIEW - Should this search be cached somwhow?
```
sharpedavid, line 232: I'd rather not until performance becomes a problem. I don't want to have to deal with cache invalidation too early.

sharpedavid (changes requested):
Initial review complete. Some feedback. I did a bit of manual testing, and it seems be be behaving nicely.

frontend/src/components/UserSearch.vue
```diff
@@ -187,12 +221,25 @@ export default {
     },
     searchUser: function(queryParameters) {
       this.userSearchLoadingStatus = true;
+      let roleId = this.roleInput.trim();
 
       UsersRepository.get(
         "?briefRepresentation=false&first=0&max=300" + queryParameters
```
sharpedavid, line 227: The 300 limit on search results might be a significant issue for this role search feature. The console currently:

1. Runs a search uses all the query parameters except for role (limit 300).
1. Runs a separate search to get all users that have the specified role (with no limit?).
1. Filters the first search result set using the second.

The problem is that if there are 1000 users with role "GIS - Admin", the search will only ever return 300 at best.

backend/src/test/java/ca/bc/gov/hlth/mohums/integration/MoHUmsIntegrationTests.java
gusslima: Got it. That change slipped through the commit I made :-/

frontend/src/components/UserSearch.vue
gusslima: Yes, it needs an override of `toString()` but it is defined to store the Client ID and Role Name required to query the users against the UMS API. I have been reading the Vue documentation but it talks mostly about Components, which might not be the case here. Anyways, it seems like the entire Searcy By Role might need to be reworked a bit due to the current needs to support more advanced queries.

sharpedavid: > Anyways, it seems like the entire Searcy By Role might need to be reworked a bit due to the current needs to support more advanced queries.

You're right, but I think this Pull Request is still heading in the right direction. After that meeting we had, it seems we'll need to add:

* Ability to select multiple roles
* Ability to choose which columns to display

But we needn't add that in this PR. We can finish-up this PR's remaining issues (if we can), merge it, and carry on later. If the code was outright broken I'd be reluctant to merge it into master, but as is it's just missing needed features that we may not have tried to do all at once anyhow.

frontend/src/components/UserSearch.vue
gusslima: Added empty entry at the beginning of the Role dropdown to clear the value of `roleInput`. The fix should address this bug.

frontend/src/components/UserSearch.vue
```diff
@@ -180,7 +180,7 @@ export default {
       ],
       organizations: organizations,
       clients: [],
-      clientRoles: [],
+      clientRoles: [ "" ],
```
sharpedavid, line 183: Does this fix the bug?

frontend/src/components/UserSearch.vue
```diff
@@ -153,18 +179,26 @@ export default {
...
       emailInput: "",
       organizationInput: "",
+      roleInput: "",
       searchResults: [],
       userSearchLoadingStatus: false,
       advancedSearchSelected: false
     };
   },
+  async created() {
+    // TODO error handling
```
sharpedavid, line 198: Is error handling still needed here? It's OK to throw exceptions for the most part, unless you have some particular error case you want to cover.

frontend/src/components/UserSearch.vue
sharpedavid: Nevermind, I can see you already answered this.

PR #59, opened 2021-02-11
Ab#2225 user update loading indicator

sharpedavid:
https://dev.azure.com/cyrovalente/IAM%20Project/_workitems/edit/2225/

A loading screen is shown after selecting a user in the search results. Previously the content area would be completely blank until the content was loaded. The loading screen looks like this. There is also a nice animation.

image

The structure of the grey boxes, the "skeleton", is actually determined by the `type="article, button, article"` attribute on the component. Lots of different arrangements are possible: https://v2.vuetifyjs.com/en/components/skeleton-loaders/. I chose this one because it kind of corresponds to the page layout: content, button, content. In the layout I've specified the button is misaligned, but I can't change that without delving into the CSS.

### DIFF
--- a/frontend/src/components/UserUpdate.vue
+++ b/frontend/src/components/UserUpdate.vue
@@ -1,12 +1,24 @@
 <!--suppress XmlInvalidId -->
 <template>
-  <div id="user-info" v-show="user.username">
-    <h1>Update - {{ user.username }}</h1>
-    <user-details :userId="this.$route.params.userid" ref="userDetails">
-      <v-btn id="submit-button" class="secondary" medium v-on:click.prevent="updateUser">Update User</v-btn>
-    </user-details>
-    <user-update-roles :userId="this.$route.params.userid"></user-update-roles>
-    <user-update-groups :userId="this.$route.params.userid"></user-update-groups>
+  <div>
+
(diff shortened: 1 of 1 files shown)


frontend/src/components/UserUpdate.vue
```diff
@@ -1,12 +1,23 @@
...
+  <div>
+    <v-skeleton-loader
+        :loading="loading"
+        v-show="!user.username"
+        ref="skeleton"
+        :boilerplate="boilerplate"
+        tile="true"
+        type="article, button, article"
+    >
+    </v-skeleton-loader>
```
sharpedavid, line 12: The content below line 12 hasn't actually changed, just the indentation.

frontend/src/components/UserUpdate.vue
```diff
@@ -24,6 +35,12 @@ export default {
     UserDetails,
     UserUpdateRoles
   },
+  data() {
+    return {
+      loading: true,
+      boilerplate: false,
```
sharpedavid, line 41: I tried to use these values directly in the attribute instead of binding them, but it broke the animation. I don't know why.

tschia (approved):

PR #60, opened 2021-03-04
AB#2224 Add download CSV button to search results page.

sharpedavid:
"Download Results" button only visible when results > 0:

image

CSV looks like this:

image

### DIFF
--- a/frontend/src/components/UserSearch.vue
+++ b/frontend/src/components/UserSearch.vue
@@ -132,7 +132,19 @@
           :loading="userSearchLoadingStatus"
           loading-text="Searching for users"
           v-on:click:row="selectUser"
-        ></v-data-table>
+        >
+          <template v-if="searchResults.length > 0" v-slot:footer>
+            <v-toolbar flat>
+              <v-spacer/>
+              <download-csv
+                  :data="searchResults"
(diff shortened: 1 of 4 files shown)


frontend/src/components/UserSearch.vue
```diff
@@ -132,7 +132,19 @@
           :loading="userSearchLoadingStatus"
           loading-text="Searching for users"
           v-on:click:row="selectUser"
-        ></v-data-table>
+        >
+          <template v-if="searchResults.length > 0" v-slot:footer>
+            <v-toolbar flat>
+              <v-spacer/>
```
sharpedavid, line 138: This template, toolbar, spacer stuff is just for a bit of layout. The `download-csv` element is all that matters.

frontend/src/components/UserSearch.vue
```diff
@@ -132,7 +132,19 @@
...
+              <download-csv
+                  :data="searchResults"
+                  :fields="['id', 'username', 'enabled', 'firstName', 'lastName', 'email']"
```
sharpedavid, line 141: I think this will be easy to work with and extend. If we need roles, we just need to get that data into the `searchResults` and figure out how to make `fields` dynamic.

Alternatively, make a copy of `searchResults` and modify it however we want, then don't use `fields` at all.

gusslima (approved):

PR #61, opened 2021-03-11
Ab#2575 advanced search multi role

gusslima:
Preliminary solution for Search By Multiple Role.
Please note that multiple roles are supported only from the same application. The current UI solution does not support multiple roles from multiple applications at the same time.

### DIFF
--- a/backend/src/main/resources/application.yaml
+++ b/backend/src/main/resources/application.yaml
@@ -47,8 +47,9 @@ server:
 #    key-store-type: pkcs12
 #    key-alias: localhost
 #    key-password: password
-#logging:
-#  level:
+logging:
+  level:
 #    KC_WEB_CLIENT: DEBUG
+    LOAD_TEST: DEBUG
 #    org.apache.coyote.http11.Http11InputBuffer: DEBUG
 #    root: DEBUG
--- a/backend/src/main/resources/application.yaml
+++ b/backend/src/main/resources/application.yaml
@@ -47,9 +47,9 @@ server:
 #    key-store-type: pkcs12
 #    key-alias: localhost
 #    key-password: password
-logging:
-  level:
+#logging:
+#  level:
 #    KC_WEB_CLIENT: DEBUG
-    LOAD_TEST: DEBUG
+#    LOAD_TEST: DEBUG
 #    org.apache.coyote.http11.Http11InputBuffer: DEBUG
 #    root: DEBUG
--- a/frontend/src/components/UserSearch.vue
+++ b/frontend/src/components/UserSearch.vue
@@ -246,7 +246,7 @@ export default {
         && this.selectedRoles.length > 0;
       
       UsersRepository.get(
-        "?briefRepresentation=false&first=0&max=300" + queryParameters
+        "?briefRepresentation=false&first=0&max=10000" + queryParameters
       )
         .then(response => {
           let results = response.data;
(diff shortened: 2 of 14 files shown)


backend/src/main/resources/application.yaml
```diff
@@ -47,8 +47,9 @@ server:
 #    key-store-type: pkcs12
 #    key-alias: localhost
 #    key-password: password
-#logging:
-#  level:
+logging:
+  level:
 #    KC_WEB_CLIENT: DEBUG
+    LOAD_TEST: DEBUG
```
sharpedavid, line 53: Don't forget to turn this off.

frontend/src/components/UserSearch.vue
```diff
@@ -216,68 +245,105 @@ export default {
       this.$router.push({ name: "UserCreate" });
     },
     searchUser: function(queryParameters) {
```
sharpedavid, line 247: I noticed that errors and warnings don't go away on subsequent searches:

image

This isn't a bug caused by any of your code, but certainly I'm noticing it now that we have the "max search results" warning. I think errors are cleared by calling this method at the start of the function, so please confirm and add:

```javascript
this.$store.commit("alert/dismissAlert");
```

frontend/src/components/UserSearch.vue
```diff
@@ -216,68 +245,105 @@ export default {
...
         })
         .catch(error => {
           this.handleError("User search failed", error);
         })
-        .finally(() => (this.userSearchLoadingStatus = false));
+        .finally(() => {
+          if (!isSearchByRole) {
+            this.userSearchLoadingStatus = false;
+          }
+        });
```
sharpedavid, line 276: I am not a Javascript expert, so please beware of what I'm about to propose. I think this code can be cleaned up using `async/await` instead of Promises. In my quick test, the following code works:

```javascript
searchUser: async function(queryParameters) {
  this.$store.commit("alert/dismissAlert");
  const maxSearch = app_config.config.max_search
                      ? app_config.config.max_search
                      : (this.maxResults * 10);
  this.userSearchLoadingStatus = true;
  let isSearchByRole = this.advancedSearchSelected
    && this.selectedRoles.length > 0;
  try {
    let results = (await UsersRepository.get(`?briefRepresentation=false&first=0&max=${maxSearch}` + queryParameters)).data;
    if (isSearchByRole) {
      results = await this.filterUsersByRole(results, maxSearch);
    }
    this.setSearchResults(results);
  } catch (e) {
    this.handleError("User search failed", e);
  } finally {
    this.userSearchLoadingStatus = false;
  }
}
```

While the code you've written is short enough that it's easy to understand, please consider this alternative.

sharpedavid:
I briefly tested this and it seems to be working well.

I am concerned that the client will want us to search by multiple clients and multiple roles at the same time. Based on our meeting, it seems that they search for users with roles in Application A and/or Application B. If the account is inactive, they remove the role for Application A. If the user does not have a role in Application B, the account is disabled. If the user does have a role in Application B, the account is not disabled.

I'm not asking for any changes to the implementation, let's stick with this, but I anticipate changes are coming.

backend/src/main/resources/application.yaml
gusslima: Addressed by https://github.com/bcgov/moh-keycloak-user-management/pull/61/commits/72b2a37b2520df71ff527fcfdd4341c46084eac5

frontend/src/components/UserSearch.vue
gusslima: Fixed by https://github.com/bcgov/moh-keycloak-user-management/pull/61/commits/f80cefb45f4fc09a273ef5d12519b050f21e1e52

frontend/src/components/UserSearch.vue
gusslima: Addressed by https://github.com/bcgov/moh-keycloak-user-management/pull/61/commits/f80cefb45f4fc09a273ef5d12519b050f21e1e52

PR #65, opened 2021-04-20
AB#2846 Search results can be opened in a new tab.

sharpedavid:
Search results can now be opened in a new tab by clicking the Username in the Search Results. The icon I used to indicate this functionality is a classic icon used by Outlook, Gmail, and other programs to indicate that the item will be opened in a new window or tab. 

image

Also fix bug in Basic Search. Pressing "Enter" key to run search wasn't working.

Also save UserSearch component during Vue router navigation to restore search results on Back button.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 .DS_Store
 node_modules
-/dist
+/frontend/dist/
 
 # local env files
 .env.local
--- a/frontend/src/components/UserSearch.vue
+++ b/frontend/src/components/UserSearch.vue
@@ -23,7 +23,7 @@
           dense
           v-model="userSearchInput"
           placeholder="Username, email, name, or ID"
-          @keyup.native.enter="searchUser"
+          @keyup.enter="searchUser('&search='+userSearchInput)"
         />
       </v-col>
       <v-col class="col-4">
@@ -171,6 +171,13 @@
           loading-text="Searching for users"
           v-on:click:row="selectUser"
         >
+          <!-- https://stackoverflow.com/questions/61394522/add-hyperlink-in-v-data-table-vuetify -->
+          <template #item.username="{ item }">
+            <a target="_blank" :href="`#/users/${item.id}`" v-on:click="openNewTab">
+              {{ item.username }}
+            </a>
+            <v-icon small>mdi-open-in-new</v-icon>
+          </template>
           <template v-if="searchResults.length > 0" v-slot:footer>
             <v-toolbar flat>
               <v-spacer/>
@@ -221,7 +228,8 @@ export default {
       organizationInput: "",
       searchResults: [],
       userSearchLoadingStatus: false,
-      advancedSearchSelected: false
+      advancedSearchSelected: false,
+      newTab: false
     };
   },
   async created() {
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -1,6 +1,6 @@
 .DS_Store
 node_modules
-/dist
+/frontend/dist/
```
sharpedavid, line 3: We forgot to update the gitignore after moving `dist` to `frontend` maybe, and now the `dist` folder has been added. I'll remove it from `master` later.

frontend/src/components/UserSearch.vue
```diff
@@ -23,7 +23,7 @@
           dense
           v-model="userSearchInput"
           placeholder="Username, email, name, or ID"
-          @keyup.native.enter="searchUser"
+          @keyup.enter="searchUser('&search='+userSearchInput)"
```
sharpedavid, line 26: This fixes a bug where the Basic Search was failing when you press "Enter" in the text field. The method call isn't new, it's actually the same method call when you click the Search button:
```html
<v-btn id="search-button" class="secondary" medium @click.native="searchUser('&search='+userSearchInput)">
```

I think Gus just missed updating it when he was adding role search, or something.

frontend/src/components/UserSearch.vue
```diff
@@ -171,6 +171,13 @@
           loading-text="Searching for users"
           v-on:click:row="selectUser"
         >
+          <!-- https://stackoverflow.com/questions/61394522/add-hyperlink-in-v-data-table-vuetify -->
+          <template #item.username="{ item }">
+            <a target="_blank" :href="`#/users/${item.id}`" v-on:click="openNewTab">
+              {{ item.username }}
+            </a>
+            <v-icon small>mdi-open-in-new</v-icon>
+          </template>
```
sharpedavid, line 180: We customize the `v-data-table` by using the "item slot". This renders the item as an `href` link instead of the default text. (I find this pretty crazy stuff. We get away from PrimeFaces and JSF for the sweet purity of HTML and Javascript, and find ourselves immediately mired in SPA's own hocus pocus.)

The `v-on:click="openNewTab"` event sets a flag to prevent the default data table's `v-on:click:row="selectUser"` from running, otherwise it opens a new tab _and_ loads the details in the same tab. I couldn't think of a better way to do this. I tried figuring out a way to determine what the user click inside the `selectUser` method, but couldn't find anything.

frontend/src/views/Users.vue
```diff
@@ -1,5 +1,7 @@
 <template>
   <div >
+    <keep-alive include="UserSearch">
       <router-view></router-view>
+    </keep-alive>
```
sharpedavid, line 5: This keeps the "UserSearch" component "alive" when routing between Views. This means that you can open a User's Details, make an edit, and then click the browser's "Back" button to return to your search results: it even saves the form data.

Normally I try to keep my Pull Requests as simple as possible, but this seems like related functionality and it's a one-line change for an immediate improvement.

camilleEsti: Is it the behavior we want, to save the data when using the browser Back button? 
Does it actually send the data to Keycloak? or does it just save it in the browser?

tschia (approved):

frontend/src/views/Users.vue
tschia: It's not persisted anywhere (since it's just search inputs) just stored in the browser cache
See: https://vuejs.org/v2/guide/components-dynamic-async.html

camilleEsti: This means that you can open a User's Details, make an edit, and then click the browser's "Back" button to return to your search results: it even saves the form data. => so that means the Search form, not the User Details form? I get it now

PR #67, opened 2021-04-30
AB#3092 Specify logback log file rotation configuration

sharpedavid:
[AB#3092](https://dev.azure.com/cyrovalente/14004b17-200a-4897-9dd5-d44e111f33a1/_workitems/edit/3092) Specify logback configuration for rotation. Also, specifying log file location ensures log file is created when application is running under `systemd`.

I just grabbed an example logback configuration from [this blog post](https://www.baeldung.com/spring-boot-logging#logback-configuration-logging). I haven't tested it much, but right now our needs are simple: we want a log file to be created, and we want it to "roll over". This example does that, it creates a log file, and it rolls it over every day or when the log reaches 10MB.

This branch is running on DEV right now, and it seems to be working:

```
[gfadmin@fireblade ums]$ ls
application.yaml  fireblade.p12  logs  user-management-service.jar
[gfadmin@fireblade ums]$ cd logs
[gfadmin@fireblade logs]$ ls
spring-boot-logger.log
```

### DIFF
--- a/backend/src/main/resources/logback-spring.xml
+++ b/backend/src/main/resources/logback-spring.xml
@@ -14,7 +14,7 @@
 
     <appender name="RollingFile"
               class="ch.qos.logback.core.rolling.RollingFileAppender">
-        <file>${LOGS}/spring-boot-logger.log</file>
+        <file>${LOGS}/ums.log</file>
         <encoder
                 class="ch.qos.logback.classic.encoder.PatternLayoutEncoder">
             <Pattern>%d %p %C{1.} [%t] %m%n</Pattern>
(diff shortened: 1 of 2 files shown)


backend/src/main/resources/logback-spring.xml
```diff
@@ -0,0 +1,44 @@
...
+        <encoder
+                class="ch.qos.logback.classic.encoder.PatternLayoutEncoder">
+            <Pattern>%d %p %C{1.} [%t] %m%n</Pattern>
+        </encoder>
+
+        <rollingPolicy
+                class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
+            <!-- http://logback.qos.ch/manual/appenders.html#TimeBasedRollingPolicy -->
+            <!-- rollover daily and when the file reaches 10 MegaBytes -->
+            <fileNamePattern>${LOGS}/archived/ums-%d{yyyy-MM-dd}.%i.log
```
sharpedavid, line 27: Interestingly, logback determines the rollover time/date by the log file name pattern. Because we've specified `yyyy-MM-dd`, it roles over daily. If we specified `yyyy-MM-dd-HH`, it would roll over hourly.

tschia (approved):

PR #69, opened 2021-05-13
removed lock status from user managment console

tschia:


ahoplock (approved):
Reviewed and approved. I haven't tried to build/run this locally, so there's that.

PR #71, opened 2021-05-19
Ab#2200 last log advanced search

camilleEsti:
Need to discuss how to improve the UI.


backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
```diff
@@ -54,14 +67,122 @@ public UsersController(WebClientService webClientService, @Value("${config.vanit
...
+            List<Object> eventsLastLog = (ArrayList) webClientService.getEvents(queryEventLastLogParams).getBody();
+            
+            Integer start = 0;
+            while(eventsLastLog != null && !eventsLastLog.isEmpty()){
+            
+                allEventsLastLog.addAll(eventsLastLog);
+                
+                if(eventsLastLog.size() == valueIteration){
+                    start = start+valueIteration;
+                    queryEventLastLogParams = buildQueryEventActiveParam(start.toString(), valueIteration.toString() , lastLogAfter, lastLogBefore);
```
sharpedavid, line 106: This block of code, from line 91 to 112, is loading all login events form the past year, right?

So on line, 96, we start at event 0, go up to 100, for one year ago.

```java
MultiValueMap<String, String> queryEventLastLogParams = buildQueryEventActiveParam("0", valueIteration.toString(), oneYearAgoParam, Optional.empty());
```

Maybe maybe there was more than 100 log events, so we need to get more:

```java
queryEventLastLogParams = buildQueryEventActiveParam(start.toString(), valueIteration.toString() , lastLogAfter, lastLogBefore);
```

I see we've updated the `start` parameter here, as expected, but why did we update the `dateFrom` and `dateTo` parameters? Shouldn't they stay the same if we're still just looking for the "one year" range?

frontend/env_config/test/keycloak.json
```diff
@@ -1,6 +1,6 @@
 {
     "realm": "moh_applications",
-    "auth-server-url": "https://common-logon-test.hlth.gov.bc.ca/auth/",
+    "auth-server-url": "https://common-logon-dev.hlth.gov.bc.ca/auth/",
```
tschia, line 3: Is this an accidental commit? pointing the test config to dev

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
camilleEsti: yes, you are right, the same event should run over and over. It's typo/copy/paste error. I'll fix it in next commit.

PR #72, opened 2021-05-20
AB#2843 Display role in search results and include in CSV download.

sharpedavid:
Story [User Management Console - Display role in advanced search result (if role was searched for)](https://dev.azure.com/cyrovalente/IAM%20Project/_sprints/taskboard/IAM%20Project%20Team/IAM%20Project/Sprint%2033?workitem=2843)

image

### DIFF
--- a/frontend/src/components/UserSearch.vue
+++ b/frontend/src/components/UserSearch.vue
@@ -1,3 +1,4 @@
+<!--suppress XmlInvalidId -->
 <template>
   <div>
     <!-- Basic Search -->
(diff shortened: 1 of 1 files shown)


frontend/src/components/UserSearch.vue
```diff
@@ -334,21 +336,43 @@ export default {
...
+                  searchResult.role = user.role;
+                }
+              }
+            }
+          }
+          let userIds = userIdsAndRoles.map(userIdAndRole => userIdAndRole.id);
+          return searchResults.filter(
+              user => userIds.includes(user.id)
+          );
+        });
```
sharpedavid, line 375: I had a hard time writing this code because Promises and stream (map, flatMap, filter) are hard to wrap my head around, but I think the code is actually fairly clean and to the point.

sharpedavid:
> 
> 
> Looks good, works well. Kind of nitpicky but it would be nice if the role column didn't show on the basic search results table.
> Could do something like this rather than have separate tables
> https://stackoverflow.com/questions/60758143/vuetify-data-table-conditional-column-rendering

Good idea and thanks for looking up a solution. I'll try it out.

PR #74, opened 2021-05-21
Code review suggestions.

sharpedavid:
I didn't find any bugs, but I did a few things to shorten the code a little:

* A `do while` loop
* A `groupingBy` Stream to collect the "Events per user".
* A `max` operation on the "Events per user" to find the last login.

### DIFF
--- a/backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
+++ b/backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
@@ -1,12 +1,15 @@
 package ca.bc.gov.hlth.mohums.controller;
 
-import ca.bc.gov.hlth.mohums.util.FilterUserByOrgId;
 import ca.bc.gov.hlth.mohums.exceptions.HttpUnauthorizedException;
 import ca.bc.gov.hlth.mohums.util.AuthorizedClientsParser;
+import ca.bc.gov.hlth.mohums.util.FilterUserByOrgId;
 import ca.bc.gov.hlth.mohums.webclient.WebClientService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.ResponseEntity;
+import org.springframework.util.CollectionUtils;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
 import org.springframework.web.bind.annotation.*;
@@ -14,18 +17,13 @@ import org.springframework.web.bind.annotation.*;
 import java.net.URI;
 import java.time.Instant;
 import java.time.LocalDate;
-import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.time.temporal.ChronoUnit;
 import java.util.*;
-import java.util.function.Predicate;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.springframework.util.CollectionUtils;
 
 @RestController
 public class UsersController {
@@ -33,14 +31,14 @@ public class UsersController {
     private final WebClientService webClientService;
 
     private final String vanityHostname;
-    
+
     private final Logger logger = LoggerFactory.getLogger(UsersController.class);
-    
+
     public UsersController(WebClientService webClientService, @Value("${config.vanity-hostname}") String vanityHostname) {
         this.webClientService = webClientService;
         this.vanityHostname = vanityHostname;
     }
-    
+
     @GetMapping("/users")
     public ResponseEntity<List<Object>> getUsers(
             @RequestParam Optional<Boolean> briefRepresentation,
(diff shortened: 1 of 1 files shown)


backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
```diff
@@ -67,116 +65,72 @@ public UsersController(WebClientService webClientService, @Value("${config.vanit
...
-                    start = start+valueIteration;
-                    queryEventLastLogParams = buildQueryEventActiveParam(start.toString(), valueIteration.toString(), oneYearAgoParam, Optional.empty());
-                    eventsLastLog = (ArrayList) webClientService.getEvents(queryEventLastLogParams).getBody();
-                } else {
-                    break;
-                }
-            }
-            logger.info("final size for AllEventsLastLog: " + allEventsLastLog.size());
+                start += maxEvents;
+            } while (!CollectionUtils.isEmpty(eventsLastLog));
```
sharpedavid, line 94: Using a `do while` here makes the code a bit shorter because we don't need to duplicate the `buildQueryEventActiveParam` and `getEvents` calls.

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
```diff
@@ -67,116 +65,72 @@ public UsersController(WebClientService webClientService, @Value("${config.vanit
...
+
+            Map<Object, List<Object>> loginEventsByUser = allEventsLastLog.stream()
+                    .filter(event -> event.get("userId") != null)
+                    .collect(Collectors.groupingBy(o -> ((LinkedHashMap) o).get("userId")));
```
sharpedavid, line 98: This Stream does what you suggested below in a comment: "better way to do this will be to build a map of Event per users". It builds a map of Events, keyed by `userId`.

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
```diff
@@ -67,116 +65,72 @@ public UsersController(WebClientService webClientService, @Value("${config.vanit
...
+                if (!userId.isEmpty()){
+
+                    if (!CollectionUtils.isEmpty(loginEventsByUser.get(userId))) {
                         LocalDate lastLogAfterDate, lastLogBeforeDate;
-                        
-                        Object userLastLogTime = ((LinkedHashMap)eventsByUserID.get(0)).get("time");
+
+                        Object userLastLogTime = loginEventsByUser.get(userId).stream()
+                                .max(Comparator.comparing(o -> (Long) ((LinkedHashMap) o).get("time")))
+                                .map(o -> ((LinkedHashMap) o).get("time")).get();
```
sharpedavid, line 110: This finds the max `time`, which should be the most recent user login. No need to convert to an actual `Instant` because we just need to find the biggest `Long`.

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
```diff
@@ -67,116 +65,72 @@ public UsersController(WebClientService webClientService, @Value("${config.vanit
...
+
                         LocalDate userLastLogLocalDate = LocalDate.ofInstant(Instant.ofEpochMilli((Long) userLastLogTime),  ZoneId.of("America/Los_Angeles"));
-                        
+                        //Or we could send the time as is to the Frontend, without formatting
+                        ((LinkedHashMap) user).put("lastLogDate", userLastLogLocalDate.format(DateTimeFormatter.ofPattern("yyyy-MM-dd")));
```
sharpedavid, line 114: I think we should just send it to the Frontend without formatting. That's what the `Events` API behaves.

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
```diff
@@ -1,46 +1,44 @@
...
-import org.springframework.util.CollectionUtils;
 
 @RestController
 public class UsersController {
 
     private final WebClientService webClientService;
 
     private final String vanityHostname;
-    
+
```
sharpedavid, line 34: If you're seeing this whitespace change, go ahead and hide it:

image

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
```diff
@@ -333,14 +287,14 @@ boolean isAuthorizedToViewClient(String token, String clientGuid) {
...
-            queryEventParams.add("max", nbElementMax);
-            dateFrom.ifPresent(dateFromValue -> queryEventParams.add("dateFrom", dateFromValue));
-            dateTo.ifPresent(dateToValue -> queryEventParams.add("dateTo", dateToValue));
-            return queryEventParams;
+    private MultiValueMap<String, String> buildQueryEventActiveParam (int start, int nbElementMax, Optional<String> dateFrom, Optional<String> dateTo){
+
+        MultiValueMap<String, String> queryEventParams = new LinkedMultiValueMap<>();
+        queryEventParams.add("type", "LOGIN");
+        queryEventParams.add("first", String.valueOf(start));
+        queryEventParams.add("max", String.valueOf(nbElementMax));
```
sharpedavid, line 295: In the main function body we want to deal with `int`, so I push the conversion down to the code that actually cares, which is setting the query parameters.

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
camilleEsti: that looks great, I'll just add back the condition on the number of events returned

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
camilleEsti: Looks better

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
camilleEsti: Thanks, looks easier to read

PR #77, opened 2021-06-14
#3347 sfds mailbox permissions

tschia:
This branch adds the ability to add SFDS mailbox authorizations for the SFDS client
The list of mailboxes and uses is currently just a made up list. In the future this will query the SFDS web service to get the lists. 

merge into develop for a future release

### DIFF
--- a/frontend/public/config.json
+++ b/frontend/public/config.json
@@ -1,6 +1,6 @@
 {
-  "max_search": 300,
-  "max_results": 30,
+  "max_search": 5000,
+  "max_results": 5000,
   "siteminder_logout": "https://logontest7.gov.bc.ca/clp-cgi/logoff.cgi?retnow=1&returl=localhost:8080",
   "service_url": "http://localhost:9090"
 }
--- a/frontend/src/components/UserCreate.vue
+++ b/frontend/src/components/UserCreate.vue
@@ -3,7 +3,7 @@
   <div id="user">
     <h1>Create New User</h1>
     <user-details ref="userDetails">
-      <v-btn id="submit-button" class="secondary" medium @click="createUser()">Create User</v-btn>
+      <v-btn id="submit-button" class="primary" medium @click="createUser()">Create User</v-btn>
     </user-details>
   </div>
 </template>
--- a/frontend/src/components/UserCreate.vue
+++ b/frontend/src/components/UserCreate.vue
@@ -2,9 +2,7 @@
 <template>
   <div id="user">
     <h1>Create New User</h1>
-    <user-details ref="userDetails">
-      <v-btn id="submit-button" class="primary" medium @click="createUser()">Create User</v-btn>
-    </user-details>
+    <user-details ref="userDetails" update-or-create="Create" @submit-user-updates="createUser"></user-details>
   </div>
 </template>
 
@@ -19,7 +17,7 @@ export default {
     UserDetails
   },
   methods: {
-    createUser: function() {
+    createUser: function(userDetails) {
       if (!this.$refs.userDetails.$refs.form.validate()) {
         this.$store.commit("alert/setAlert", {
           message: "Please correct errors before submitting",
@@ -28,7 +26,8 @@ export default {
         window.scrollTo(0, 0);
         return;
       }
-      UsersRepository.createUser(this.user)
+
+      UsersRepository.createUser(userDetails)
         .then(response => {
           //Keycloak returns the newly created user id in the response location
           let responseLocation = response.headers.location;
@@ -56,11 +55,6 @@ export default {
         .finally(() => {
           window.scrollTo(0, 0);
         });
-    }
-  },
-  computed: {
-    user() {
-      return this.$store.state.user;
     }
   }
 };
--- a/frontend/src/components/UserCreate.vue
+++ b/frontend/src/components/UserCreate.vue
@@ -18,15 +18,6 @@ export default {
   },
   methods: {
     createUser: function(userDetails) {
-      if (!this.$refs.userDetails.$refs.form.validate()) {
-        this.$store.commit("alert/setAlert", {
-          message: "Please correct errors before submitting",
-          type: "error"
-        });
-        window.scrollTo(0, 0);
-        return;
-      }
-
       UsersRepository.createUser(userDetails)
         .then(response => {
           //Keycloak returns the newly created user id in the response location
--- a/frontend/src/components/UserDetails.vue
+++ b/frontend/src/components/UserDetails.vue
@@ -148,6 +148,15 @@ export default {
         });
     },
     updateUser: function() {
+      // Validate the User Details
+      if (!this.$refs.form.validate()) {
+        this.$store.commit("alert/setAlert", {
+          message: "Please correct errors before submitting",
+          type: "error"
+        });
+        window.scrollTo(0, 0);
+        return;
+      }
       this.$emit('submit-user-updates', this.user)
     }
   },
--- a/frontend/src/components/UserDetails.vue
+++ b/frontend/src/components/UserDetails.vue
@@ -90,7 +90,7 @@
           </ul>
         </v-col>
       </v-row>
-      <v-btn id="submit-button" class="primary" medium @click="updateUser">Update User</v-btn>
+      <v-btn id="submit-button" class="primary" medium @click="updateUser">{{ updateOrCreate }} User</v-btn>
     </v-card>
   </div>
 </template>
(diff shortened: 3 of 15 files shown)


frontend/public/config.json
```diff
@@ -1,6 +1,6 @@
 {
-  "max_search": 300,
-  "max_results": 30,
+  "max_search": 5000,
+  "max_results": 5000,
```
tschia, line 3: Just upped these for local development

frontend/src/components/UserCreate.vue
```diff
@@ -2,9 +2,7 @@
 <template>
   <div id="user">
     <h1>Create New User</h1>
-    <user-details ref="userDetails">
-      <v-btn id="submit-button" class="secondary" medium @click="createUser()">Create User</v-btn>
-    </user-details>
+    <user-details ref="userDetails" update-or-create="Create" @submit-user-updates="createUser"></user-details>
```
tschia, line 5: The user to update is now emitted from the child component (user-details) rather than pulled from the vuex store.

frontend/src/components/UserDetails.vue
```diff
@@ -112,7 +112,19 @@ export default {
...
+        firstName: '',
+        lastName: '',
+        email: '',
+        attributes: {
+          phone: null,
+          org_details: null,
+          access_team_notes: null
+        },
+        federatedIdentities: null
+      }
```
tschia, line 127: We now use a local user model instead of working directly off of the vuex store

frontend/src/components/UserDetails.vue
```diff
@@ -127,10 +139,25 @@ export default {
...
+      // Validate the User Details
+      if (!this.$refs.form.validate()) {
+        this.$store.commit("alert/setAlert", {
+          message: "Please correct errors before submitting",
+          type: "error"
+        });
+        window.scrollTo(0, 0);
+        return;
+      }
+      this.$emit('submit-user-updates', this.user)
```
tschia, line 160: moved the user details form validation to this method instead of being duplicated in the parent components (update/create user)
this method now emits the user to update or create to the parent components

frontend/src/components/UserDetails.vue
```diff
@@ -127,10 +139,25 @@ export default {
       return UsersRepository.getUser(this.userId)
         .then(response => {
           this.user = response.data;
+          // Keycloak returns attributes as arrays which doesn't work with the autocomplete for org details
+          this.user.attributes.org_details = formatOrganization(this.user.attributes.org_details);
```
tschia, line 143: we were previously doing this in the getter for the vuex state, just moved it here.

frontend/src/components/UserSearch.vue
```diff
@@ -28,7 +28,7 @@
         />
       </v-col>
       <v-col class="col-4">
-          <v-btn id="search-button" class="secondary" medium @click.native="searchUser('&search='+userSearchInput)">Search Users</v-btn>
+          <v-btn id="search-button" class="primary" medium @click.native="searchUser('&search='+userSearchInput)">Search Users</v-btn>
```
tschia, line 31: udpated button colour to match bcgov standards 
https://developer.gov.bc.ca/Design-System/Primary-Button

frontend/src/components/UserUpdateRoles.vue
```diff
@@ -81,24 +81,31 @@
...
 
 export default {
   name: "UserUpdateRoles",
+  components: { UserUpdateSfdsAuth },
   props: ['userId'],
   data() {
     return {
       clients: [],
-      selectedClientId: null,
+      selectedClient: null,
```
tschia, line 108: map the actual client object rather than just the ID from the autocomplete menu
this lets us check the name to see if it matches the SFDS client to display the additional component

frontend/src/store/modules/user.js
```diff
@@ -25,14 +28,28 @@ const mutations = {
...
+        state.username = user.username
+        state.firstName = user.firstName
+        state.lastName = user.lastName
+        state.email = user.email
+        if (user.attributes) {
+            state.attributes.phone = user.attributes.phone
+            state.attributes.org_details = user.attributes.org_details
+            state.attributes.access_team_notes = user.attributes.access_team_notes
+        }
+    },
```
tschia, line 52: add the ability to just store the values from the user-details component

frontend/src/components/UserUpdateSfdsAuth.vue
```diff
@@ -0,0 +1,283 @@
...
+      sfdsTableHeaders: [
+        { text: 'Mailbox', value: 'm' },
+        { text: 'Uses', value: 'u' },
+        { text: 'Permission', value: 'p' },
+        { text: 'Actions', value: 'actions', sortable: false }
+      ],
+      sfdsAuthorizations: [],
+      sfdsAuthorizationsToSet: [],
+      sfdsMailboxes: ['HSCIS', 'BCMA', 'PHC', 'HOOPC'],
+      sfdsUses: ['bcma', 'hscis', 'phc', 'wda', 'hoopc', 'grp'],
```
tschia, line 131: These are just hardcoded for now. Will load from SFDS web service at a later time

frontend/src/components/UserUpdateSfdsAuth.vue
```diff
@@ -0,0 +1,283 @@
...
+import UsersRepository from "@/api/UsersRepository";
+
+export default {
+  name: "UserUpdateSfdsAuth",
+  data() {
+    return {
+      sfdsTableHeaders: [
+        { text: 'Mailbox', value: 'm' },
+        { text: 'Uses', value: 'u' },
+        { text: 'Permission', value: 'p' },
```
tschia, line 125: short hand id's for the sfds object to save space in Keycloak's limited attribute size fields

frontend/src/components/UserUpdateSfdsAuth.vue
ahoplock: Even though they are hardcoded, can you make the sfdsMailboxes lowercase?

frontend/src/components/UserUpdateSfdsAuth.vue
```diff
@@ -0,0 +1,283 @@
...
+        { text: 'Actions', value: 'actions', sortable: false }
+      ],
+      sfdsAuthorizations: [],
+      sfdsAuthorizationsToSet: [],
+      sfdsMailboxes: ['HSCIS', 'BCMA', 'PHC', 'HOOPC'],
+      sfdsUses: ['bcma', 'hscis', 'phc', 'wda', 'hoopc', 'grp'],
+      sfdsPermissions: ['get', 'send', 'get-send', 'get-delete', 'get-send-delete'],
```
ahoplock, line 132: ...you already picked up on this in the demo but I think Vlad has permissions pared down to a single number.

ahoplock (commented):
Left a couple comments, otherwise reviewed and approved from what I see in github. I haven't run this locally to test it out though, so I'll leave that for either @sharpedavid or @camilleEsti when they get around to reviewing this.

frontend/src/components/UserCreate.vue
sharpedavid: Is the `update-or-create` attribute used? I don't see anything that references it.

frontend/src/components/UserUpdateSfdsAuth.vue
```diff
@@ -0,0 +1,283 @@
...
+      currentSfdsAuthorization: {},
+      editDialog: false,
+      deleteDialog: false,
+      editedIndex: -1,
+      defaultAuthorization: {
+        m: '',
+        u: [],
+        p: ''
```
sharpedavid, line 140: Might as well use full variable names here?

frontend/src/components/UserUpdateSfdsAuth.vue
```diff
@@ -0,0 +1,283 @@
...
+          rules.push(rule2);
+        }
+      }
+      return rules;
+    }
+  },
+  methods: {
+    editItem: function (item) {
+      this.editedIndex = this.sfdsAuthorizations.indexOf(item)
+      this.currentSfdsAuthorization = Object.assign({}, item)
```
sharpedavid, line 205: Weird method, `Object.assign`. Zero way to guess what's happening: just have to know what `item`'s properties are copied into `{}` and returned to be assigned to `this.currentSfdsAuthorization`. You could also have written `Object.assign(this.currentSfdsAuthorization, item)` to the same effect. Not a suggestion, just thinking through it.

frontend/src/components/UserUpdateSfdsAuth.vue
tschia: how they're defined there is how they map into Keycloak (and this saves space in Keycloak). I think I'd need to add some kind of transformer on the object before sending it if I change the variable names.

sharpedavid: I'd just leave it. I don't care much about saving space unless we have a _really_ practical reason to, but I wouldn't bother changing it now.

sharpedavid (approved):

PR #78, opened 2021-07-13
Fix unit tests

camilleEsti:
Just replace the search by user ID with a search by username


frontend/src/loadconfig/index.js
```diff
@@ -1,5 +1,7 @@
+//import fetch from 'cross-fetch';
 let app_config = new Object();
 
+//const fetch = require("node-fetch");
```
sharpedavid, line 4: Please remove commented-out code.

frontend/tests/e2e/alltests.js
```diff
@@ -32,8 +33,8 @@ test('Test search', async t => {
     await t.typeText('#user-search', 'testcafe')
         .click('#search-button')
         .expect(Selector('html').textContent)
-        // This is the ID of the testcafe user.
-        .contains(TEST_CAFE_USER_ID);
+        // Search with Username not with ID anymore
+        .contains(TEST_CAFE_USERNAME);
```
sharpedavid, line 37: The comment "Search with Username not with ID anymore" won't be helpful in the near future because future maintainers won't know we ever searched by ID. 

"Search for the username" would be fine, or better yet no comment at all. The original comment "This is the ID of the testcafe user" was written before the TestCafe user ID was turned into a named constant (TEST_CAFE_USER_ID), but now that it is a constant the code is self-documenting.

sharpedavid (changes requested):
I ran the tests and they pass now. The "advancedSearch" tests are very error-prone though, it seems it never find "MOH_USER" in the dropdown. That has nothing to do with this fix.

Otherwise I recommended deleting (or editing) the comments. There was a couple commented-out lines of code to delete, but I did that myself in the browser editor.

PR #81, opened 2021-08-05
Ab#3368 dropdowns sfds uses

camilleEsti:
pull to Develop branch


frontend/src/components/UserUpdateSfdsAuth.vue
```diff
@@ -145,8 +150,12 @@ export default {
     };
   },
   async created() {
+    //TODO error handling
```
ahoplock, line 153: I guess this TODO is outside of the scope of this change? It'll be a rare but outages to the SFDS application would bring down the API, so how would the UMC behave if this were the case?

ahoplock (approved):

ahoplock (approved):
Reviewed and approved (I have not run the tests; nor built and deployed locally).

frontend/src/components/UserUpdateSfdsAuth.vue
camilleEsti: Errors are actually handled, caught and logged by the getUses and getMailbox methods

PR #97, opened 2021-11-22
Feature user search by app

gperkinscgi:
Filter the search by last login to take into account the selected application.

### DIFF
--- a/backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
+++ b/backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
@@ -76,37 +76,8 @@ public class UsersController {
         
         //Filter based on selected client & roles
         if (clientId.isPresent()){
-            List<Object> filteredUsers = new ArrayList<>();
-            String[] roles = null;
-            Map<String,String> clientsInRoles = new HashMap<>();
-            if (selectedRoles.isEmpty()){
-                //If no roles selected, grab all roles for the selected client
-                ResponseEntity res = webClientService.getClientRoles(clientId.get());
-                List<Map> allRoles = (List)res.getBody();
-                roles = new String[allRoles.size()];
-                for (int i=0; i<allRoles.size(); i++){
-                    roles[i] = (String)allRoles.get(i).get("name");
-                }
-            }else{
-                roles = selectedRoles.get();
-            }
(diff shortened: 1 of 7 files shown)


sharpedavid:
The search works fine, but it behaves a bit differently than the existing "search by application role" feature which may create some confusion:

* Old: Search by application only works if you specify roles.
* New: Search by application only works if you specify roles, unless you specify a last log-in date, in which case search by application works without specifying roles.

To make the search behaviour consistent, we'd have to change either the frontend of backend:

* To make it consistent on the backend, we'd have to filter by application only when roles are specify.
* Or to make it consistent on the frontend, we'd have to filter by application even if no roles were selected.

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
```diff
@@ -71,6 +73,41 @@ public UsersController(WebClientService webClientService, @Value("${config.vanit
 
             searchResults = ResponseEntity.status(searchResults.getStatusCode()).body(filteredUsers);
         }
+        
+        //Filter based on selected client & roles
```
sharpedavid, line 77: I know this method was already too long before you added this, but it's getting pretty long. Considering moving this into something like:
```java
// Filter based on selected client & roles
if (clientId.isPresent()){
    users = filterUsersByRole(clientId, selectedRoles, users);
    searchResults = ResponseEntity.status(searchResults.getStatusCode()).body(users);
}
// ...

private List<Object> filterUsersByRole(Optional<String> clientId, Optional<String[]> selectedRoles, List<Object> users)
```

frontend/src/components/UserSearch.vue
```diff
@@ -302,12 +302,25 @@ export default {
...
         });
       }
+      if (this.selectedRoles){
+        //console.log(this.selectedRoles);
+        let roles = "";
+        this.selectedRoles.forEach(selectedRole => {
+          if (roles.length>0){
+            roles = roles+",";
+          }          
+          roles = roles+selectedRole.name;
```
sharpedavid, line 319: Consider `roles = this.selectedRoles.map(role => role.name).join(",")`.

sharpedavid (changes requested):

sharpedavid (approved):

PR #98, opened 2021-11-23
Feature user details show all roles

gperkinscgi:
Added read-only list of roles for all clients

### DIFF
--- a/frontend/src/components/UserDetails.vue
+++ b/frontend/src/components/UserDetails.vue
@@ -93,6 +93,13 @@
               {{ identity.identityProvider | formatIdentityProvider }} [{{ identity.userName }}]
             </li>
           </ul>
+          <br/><br/>
+          <label for="all-roles">Application Roles</label>
+          <ul id="all-roles" style="margin-top: 5px; list-style: square">
+              <li v-for="client in user.allRoles" :key="client.clientName">
+                {{client.clientName}} [<span v-for="role in client.effectiveRoles" :key="role.id">{{role.name}} </span>]
+              </li>
+          </ul>
         </v-col>
       </v-row>
       <v-btn id="submit-button" v-if="editUserDetailsPermission" class="primary" medium @click="updateUser">{{ updateOrCreate }} User</v-btn>
@@ -102,7 +109,8 @@
 
 <script>
 import UsersRepository from "@/api/UsersRepository";
-import organizations from "@/assets/organizations"
+import organizations from "@/assets/organizations";
+import clients from "@/api/ClientsRepository";
 
 export default {
   name: "UserDetails",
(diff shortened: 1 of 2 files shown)


frontend/src/components/UserDetails.vue
```diff
@@ -173,13 +185,45 @@ export default {
...
+                    if (nameA < nameB) {
+                      return -1;
+                    }
+                    if (nameA > nameB) {
+                      return 1;
+                    }
+                    return 0;
+                });
+                //Need to force an update as vue doesn't notice the collection has changed
+                this.$forceUpdate();
```
sharpedavid, line 221: Vue is (understandably?) fussy about how it reacts to state change. Apparently it's more developer friendly in Vue 3. I think you're bumping into the complexity around objects and arrays, see this blog post and the official documentation:

* https://vuejs.org/v2/guide/reactivity.html#Change-Detection-Caveats
* https://michaelnthiessen.com/debugging-guide-why-your-component-isnt-updating/#2-make-sure-to-update-arrays-and-objects-correctly-only-in-vue-2

I recommend making `allRoles`  a root-level `data` property, and then you won't need to call `this.$forceUpdate();`

frontend/src/components/UserDetails.vue
```diff
@@ -173,13 +185,45 @@ export default {
...
+      clients.get().then(allClients => {        
+        allClients.data.forEach(client => {
+            UsersRepository.getUserEffectiveClientRoles(this.userId,client.id).then(clientRoles => {
+              if (clientRoles.data.length>0){
+                let roleCollection = {};
+                roleCollection.clientName = client.name;
+                roleCollection.effectiveRoles = clientRoles.data;
+                this.user.allRoles.push(roleCollection);
+                //Sort results by client name
+                this.user.allRoles.sort(function(a, b) {
```
sharpedavid, line 209: I think this is running N sorts, one for each client, when you could sort once at the end.

frontend/src/components/UserDetails.vue
```diff
@@ -93,6 +93,13 @@
               {{ identity.identityProvider | formatIdentityProvider }} [{{ identity.userName }}]
             </li>
           </ul>
+          <br/><br/>
+          <label for="all-roles">Application Roles</label>
+          <ul id="all-roles" style="margin-top: 5px; list-style: square">
+              <li v-for="client in user.allRoles" :key="client.clientName">
+                {{client.clientName}} [<span v-for="role in client.effectiveRoles" :key="role.id">{{role.name}} </span>]
```
sharpedavid, line 100: I was trying to figure out how to put a comma in between and came up with this:

`{{client.clientName}} [{{client.effectiveRoles.map(role => role.name).join(", ")}}]`

frontend/src/components/UserDetails.vue
```diff
@@ -129,15 +137,19 @@ export default {
...
           access_team_notes: null
         },
+        allRoles: null,
         federatedIdentities: null
       }
     };
   },
   async created() {
+    //Create global ref to allow role update from UserUpdateRoles
+    this.$root.$refs.UserDetails = this;
```
sharpedavid, line 147: I don't really know how to do better than this. My guess is something like putting the data into Vuex, but I'm really familiar with it. Maybe @tschia has a suggestion if it's worth the time.

sharpedavid (changes requested):

sharpedavid (approved):

PR #100, opened 2021-12-15
Feature hscis mailbox

gperkinscgi:
Added mailbox admin functionality to HSCIS

### DIFF
--- a/frontend/src/components/UserUpdateSfdsAuth.vue
+++ b/frontend/src/components/UserUpdateSfdsAuth.vue
@@ -176,7 +176,6 @@ export default {
     }
 
     if (sfdsAuthString !== "") {
-      console.log(sfdsAuthString);
       this.sfdsAuthorizations = JSON.parse(sfdsAuthString);
     } else {
       this.sfdsAuthorizations = [];
--- a/frontend/src/components/UserUpdateSfdsAuth.vue
+++ b/frontend/src/components/UserUpdateSfdsAuth.vue
@@ -2,11 +2,11 @@
   <v-data-table
       id="sfds-authorizations-table"
       :headers="sfdsTableHeaders"
-      :items="sfdsAuthorizations"
+      :items="filteredSfdsAuthorizations"
       class="break">
     <template v-slot:top>
       <v-toolbar flat >
-        <h2 class="sfds-authorizations-header">SFDS Authorizations</h2>
+        <h2 class="sfds-authorizations-header">Mailbox Authorizations</h2>
         <v-spacer></v-spacer>
         
         <v-dialog v-model="editDialog" max-width="840px">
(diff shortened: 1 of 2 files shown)


frontend/src/components/UserUpdateSfdsAuth.vue
```diff
@@ -158,19 +159,19 @@ export default {
     
     let sfdsAuthString = "";
     
-    if (this.$store.state.user.attributes.sfds_auth_1) {
+    if (this.$store.state.user.attributes.sfds_auth_1 && this.$store.state.user.attributes.sfds_auth_1.length>0) {
```
sharpedavid, line 162: For this number of checks, a temporary variable would make this a little easier to read. Just a personal preference, feel free to ignore.

```javascript
const attributes = this.$store.state.user.attributes;
if (attributes.sfds_auth_1 && attributes.sfds_auth_1.length > 0)
```

frontend/src/components/UserUpdateSfdsAuth.vue
```diff
@@ -158,19 +159,19 @@ export default {
...
-    if (this.$store.state.user.attributes.sfds_auth_3) {
+    if (this.$store.state.user.attributes.sfds_auth_3 && this.$store.state.user.attributes.sfds_auth_3.length>0) {
       sfdsAuthString = sfdsAuthString.concat(this.$store.state.user.attributes.sfds_auth_3[0]);
     }
-    if (this.$store.state.user.attributes.sfds_auth_4) {
+    if (this.$store.state.user.attributes.sfds_auth_4 && this.$store.state.user.attributes.sfds_auth_4.length>0) {
       sfdsAuthString = sfdsAuthString.concat(this.$store.state.user.attributes.sfds_auth_4[0]);
     }
-    if (this.$store.state.user.attributes.sfds_auth_5) {
+    if (this.$store.state.user.attributes.sfds_auth_5 && this.$store.state.user.attributes.sfds_auth_5.length>0) {
```
sharpedavid, line 174: Out of curiosity, why did we need to add the length check? What was happening without it?

frontend/src/components/UserUpdateSfdsAuth.vue
```diff
@@ -189,6 +190,12 @@ export default {
     }
   },
   computed: {
+    filteredSfdsAuthorizations() {
+      return this.sfdsAuthorizations.filter(item => this.selectedClient.clientId=='HSCIS'?item.u.includes("hscis"):!item.u.includes("hscis"));
```
sharpedavid, line 194: I'm not familiar with SFDS permissions, but I suspect the data that produced this screenshot won't be replicated in production. If a user has "combination use" record that included hscis, you would only see that record with HSCIS selected. It seems to me that after this change it won't be possible to create a combination record that includes the HSCIS use, as the HSCIS use will be selectable if and only if the HSCIS client is selected. So this is good, right @tschia ?

image

sharpedavid (approved):
I tested this manually and ran the TestCafe tests too.

PR #106, opened 2022-01-14
Ab#3927 plr env specific orgs

sharpedavid:
PLR organizations are environment-specific, so I made the `organizations.json` file environment-specific to support them. If you want to see exactly which organizations were added for PLR, look at this commit, https://github.com/bcgov/moh-keycloak-user-management/pull/106/commits/e76f26135c7178c69dc2c156a21e9f9427d587b9. I committed them separately so they would show in a diff.


frontend/src/components/UserDetails.vue
```diff
@@ -116,19 +116,19 @@
...
-    item.text = `${item.id} - ${item.name}`;
-    return item;
-  }),
   data() {
     return {
+      organizations: app_config.organizations.map((item) => {
+        item.value = JSON.stringify(item);
+        item.text = `${item.id} - ${item.name}`;
+        return item;
+      }),
```
sharpedavid, line 131: If I don't move this to the `data` section of the component, the organizations are empty. I don't really understand it, but it must be something to do with the Vue.js component life-cycle. Note that the `UserSearch` component already had `organizations` in `data`, so this change makes it consistent.

frontend/src/components/UserSearch.vue
```diff
@@ -276,7 +275,11 @@ export default {
   name: "UserSearch",
   data() {
     return {
-      organizations: organizations,
+      organizations: app_config.organizations.map((item) => {
+        item.value = JSON.stringify(item);
+        item.text = `${item.id} - ${item.name}`;
+        return item;
+      }),
```
sharpedavid, line 282: If I don't use `map`, then the organizations drop-down shows the organizations as `Object`. I copied this `map` pattern from `UserDetails`, so now it's consistent too. I don't really understand what necessitates this change when it was working before, but at least it's all consistent now.

ahoplock (approved):
Reviewed and approved; haven't tested locally, so please deploy to Dev.

PR #107, opened 2022-01-24
5454 test new builds are working 2

ahoplock:
Reduce max getAll results (10000 -> 5000) which makes clientsAuthorized pass regularly when running full test suite post-Keycloak 15.0.2 upgrade; Log only exception message instead of full stacktrace for warning

### DIFF
--- a/backend/src/main/java/ca/bc/gov/hlth/mohums/util/FilterUserByOrgId.java
+++ b/backend/src/main/java/ca/bc/gov/hlth/mohums/util/FilterUserByOrgId.java
@@ -81,7 +81,7 @@ public final class FilterUserByOrgId implements Predicate<Object> {
             // These exceptions are thrown when the value found by EXPRESSION
             // is not a valid JSON object. We can't stop people from creating
             // 'org_details' attributes that don't contain JSON data.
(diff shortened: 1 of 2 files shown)


backend/src/main/java/ca/bc/gov/hlth/mohums/util/FilterUserByOrgId.java
```diff
@@ -81,7 +81,7 @@ static String extractOrgId(Object orgDetails) {
             // These exceptions are thrown when the value found by EXPRESSION
             // is not a valid JSON object. We can't stop people from creating
             // 'org_details' attributes that don't contain JSON data.
-            LOGGER.warn("Unable to parse: \"{}\": {}.", organizationDetails, e);
```
ahoplock, line 84: This would always log a full stacktrace for a couple tests when building which makes you do a double take for no reason if you're reviewing logs.

sharpedavid: I can't recall what I found, but it would be nice if we could do this with a check instead of an exception.

PR #109, opened 2022-02-28
Develop

ahoplock:
PR for Release 1.4 to merge the latest currently in develop to master. This is primarily to release the HSCIS Mailbox changes to Prod and also includes all of the ITSB/RegAdmin enhancements we've been working on the past several months.


backend/pom.xml
```diff
@@ -10,7 +10,7 @@
     </parent>
     <groupId>ca.bc.gov.hlth</groupId>
     <artifactId>moh-ums</artifactId>
-    <version>0.0.2-SNAPSHOT</version>
+    <version>0.0.1-SNAPSHOT</version>
```
sharpedavid, line 13: Not sure what to do about this. Maybe nothing, just fix it in `master`.

ahoplock: This is what we agreed to do in the chat; leave it as 0.0.1-SNAPSHOT and just use GitHub's Releases/Tags for identifying the actual version in Prod. I think I previously had moved it from 0.0.2-SNAPSHOT to 1.3.0 or something to reflect what we were at currently but then we agreed not to use versioning in Maven at all.

PR #114, opened 2022-03-07
5947: UMC: Create User page shows loading indicator for roles forever

vangeliq:
5947: UMC: Create User page shows loading indicator for roles forever

The roles part of the page now renders only when that user already exists


ahoplock (approved):
Reviewed and approved. Haven't tested locally but will be deploying to Dev.

PR #124, opened 2022-03-30
Increased query result limit to ALL

vangeliq:
Default value for max results was 100, changed to show all results instead.

### DIFF
--- a/backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
+++ b/backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
@@ -162,7 +162,9 @@ public class UsersController {
             roles = selectedRoles.get();
         }
         for (String role:roles){
-            ResponseEntity res = webClientService.getUsersInRole(clientId,role,null);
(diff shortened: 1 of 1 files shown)


backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
```diff
@@ -162,7 +162,9 @@ private List filterUsersByRole(Optional<String[]> selectedRoles, String clientId
             roles = selectedRoles.get();
         }
         for (String role:roles){
-            ResponseEntity res = webClientService.getUsersInRole(clientId,role,null);
+            MultiValueMap<String, String> queryParam = new LinkedMultiValueMap<>();
```
ahoplock, line 165: I'd suggest calling this map variable `queryParams` even though you're currently only setting one parameter since it is a map that could hold more than one parameter (say, if we ever needed to add anything in the future).

ahoplock (approved):
Reviewed and approved. Will test when UMS is next deployed to Dev.

PR #133, opened 2022-04-26
make site title based on configuration

tschia:
also updated the configuration for each environment

### DIFF
--- a/frontend/src/components/template/TheHeader.vue
+++ b/frontend/src/components/template/TheHeader.vue
@@ -4,7 +4,7 @@
       <section class="container">
         <section class="identity">
           <img src="../../assets/images/logo.png" width="154" class="logo" alt="BC Government Logo" />
-          <div class="sitename">MoH User Management</div>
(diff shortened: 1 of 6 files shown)


frontend/src/components/template/TheHeader.vue
```diff
@@ -4,7 +4,7 @@
       <section class="container">
         <section class="identity">
           <img src="../../assets/images/logo.png" width="154" class="logo" alt="BC Government Logo" />
-          <div class="sitename">MoH User Management</div>
+          <div class="sitename">{{ title }}</div>
```
ahoplock, line 7: This is great for the `sitename` but the `<head><title>` is still MoH User Management.

ahoplock (approved):
Reviewed and approved.

ahoplock (approved):

frontend/src/components/template/TheHeader.vue
tschia: fixed

PR #134, opened 2022-05-02
4095 handdle session timeout gracefully

vangeliq:


frontend/src/api/Repository.js
```diff
@@ -12,7 +12,10 @@ function kcRequest() {
             headers: {Authorization: 'Bearer ' + keycloak.token}
         });
     }
-    return keycloak.updateToken(0).then(createAxios);
+    if(keycloak.isTokenExpired(0)) {
+        return keycloak.logout({redirectUri: app_config.config.siteminder_logout});
```
tschia, line 16: I think we can skip the SiteMinder logout in this case. (Just remove the {redirectUri: ... }

When the user clicks the logout button they expect their all of their logins to be logged out, but in the case of their session timing out that isn't necessarily true. 

Timing out on one application shouldn't necessarily cause a logout elsewhere.

ahoplock (approved):
Reviewed and approved; will test when deployed to Dev.

PR #142, opened 2022-06-14
6202 combine user permissions and management

vangeliq:


frontend/src/components/UserMailboxAuthorizations.vue
```diff
@@ -0,0 +1,94 @@
+<template>
+  <v-card  v-if="clientsWithRoles.length > 0" outlined class="subgroup">
+    <h2 class="sfds-authorizations-header">Mailbox Authorizations</h2>
+
+    <div class="nav-tabs" style="width: 100%">
+    <!-- // todo: ??????????? why does this NOT work if its selectedClient = 1?? -->
```
tschia, line 6: Is this comment still applicable?

frontend/src/components/UserMailboxAuthorizations.vue
```diff
@@ -0,0 +1,94 @@
...
+  async created() {
+    this.$root.$refs.UserMailboxAuthorizations = this;
+  },
+   watch: {
+    clients: function(){
+      this.getAllowedClients()
+}
+ },
+  // todo:why does this work if i make it outside of the created??? 
+  // todo:why does this randomly not work??
```
tschia, line 47: Are these comments still applicable

frontend/src/components/UserMailboxAuthorizations.vue
vangeliq: no - I'll delete all these comments, thanks

frontend/src/components/UserUpdate.vue
```diff
@@ -50,13 +77,24 @@ export default {
...
+      //get both SFSD and HSCIS
+      this.MailboxClients = [];
+      let clients = await ClientsRepository.get();
+      clients = clients.data;
+      clients.map((client) => {
+        if (this.MailboxClientNames.includes(client.name)) {
+          this.MailboxClients.push(client);
+        }
+      });
+    },
```
tschia, line 91: I think we could move all of this functionality into UserMailboxAuthorizations.vue
that way we can keep the names of clients with mailboxes only in mailbox related components, and don't need to pass any props between the components

frontend/src/components/UserUpdate.vue
```diff
@@ -2,33 +2,49 @@
...
     <v-skeleton-loader
-        ref="skeleton"
-        v-show="!user.username"
-        type="article, button, article"
+      ref="skeleton"
+      v-show="!user.username"
+      type="article, button, article"
+    >
+    </v-skeleton-loader
     >
```
tschia, line 10: this can be on the previous line

frontend/src/components/UserUpdate.vue
```diff
@@ -2,33 +2,49 @@
...
+      <user-update-groups
+        :userId="this.$route.params.userid"
+      ></user-update-groups>
     </div>
   </div>
 </template>
 
 <script>
-import UsersRepository from "@/api/UsersRepository";
+// import UsersRepository from "@/api/UsersRepository";
```
tschia, line 33: If this is no longer required you can just delete it

frontend/src/components/UserMailboxAuthorizations.vue
```diff
@@ -0,0 +1,101 @@
...
+export default {
+  name: "UserMailboxAuthorizations",
+  props: ["userId"],
+
+  components: {
+    UserUpdateSfdsAuth,
+  },
+  data() {
+    return {
+      MailboxClientNames: ["SFDS", "HSCIS"],
```
tschia, line 34: variable name should be camel case

frontend/src/components/UserMailboxAuthorizations.vue
```diff
@@ -0,0 +1,101 @@
...
+      clientsWithRoles: [],
+      selectedClient: 0,
+    };
+  },
+  async created() {
+    this.$root.$refs.UserMailboxAuthorizations = this;
```
tschia, line 40: What's the purpose of this line?

vangeliq: when the user role gets updated, they call UserMailBoxAuth to update & see if user has roles to view mailbox

frontend/src/components/UserUpdateRoles.vue
```diff
@@ -218,33 +392,61 @@ export default {
...
+          window.scrollTo(0, 0);})    
+    },
+  },
 };
 </script>
 
 <style>
-.right-gutters .col {
-  padding: 10px 12px 10px 0;
+.classForDialog {
```
tschia, line 422: something like updateRolesDialog might be a better name for this
I think it can go in the <style scoped> block with the rest of the css for this component

frontend/src/components/UserUpdateRoles.vue
```diff
@@ -1,210 +1,384 @@
...
-                    readonly
-                    v-for="role in effectiveClientRoles"
-                    v-model="effectiveClientRoles"
-                    :value="role"
-                    :key="role.name"
+            <!-- pop up to add something -->
+            <v-card class="popup">
+              <!-- pop up header -->
+              <v-row class="header">
+                <label for="select-client">Application</label>
```
tschia, line 25: This label can be moved into the client selector <v-row> directly above the auto complete
Adding a class with `float: right` to the `<v-row class="header">` element will make it so the X (close button) still stays in the correct spot.

frontend/src/components/UserUpdateRoles.vue
```diff
@@ -1,210 +1,384 @@
...
+                      >Save User Roles</v-btn
+                    >
+                  </div>
+                </div>
+              </div>
+            </v-card>
+          </v-dialog>
+        </v-toolbar>
+      </template>
+      <template #item.lastLogin="{ item }">
```
tschia, line 179: Can we swap the order of the lastLogin <template> block with the roleArray <template> block? 
It doesn't actually make a difference, but then their order in the file matches the order of the columns in the table itself.

frontend/src/components/UserDetails.vue
```diff
@@ -93,20 +93,6 @@
...
-          <br/><br/>
-          <label for="all-roles">Application Roles and Last Login</label>
-          <v-skeleton-loader
-              ref="roleSkeleton"
-              v-show="!rolesLoaded"
-              type="list-item@5">
-          </v-skeleton-loader>
-          <div id="user-roles" v-show="rolesLoaded">          
-            <ul id="all-roles" style="margin-top: 5px; list-style: square">
-              <li v-for="client in allRoles" :key="client.clientName">
```
tschia, line 105: The method (`loadUserRoles`) that populates `allRoles` can be removed, as well as the data variables for `allRoles` and  `rolesLoaded`

frontend/src/components/UserUpdateRoles.vue
```diff
@@ -1,210 +1,384 @@
...
-    }
+      return !!this.$keycloak.tokenParsed.resource_access[
+        umsClientId
+      ].roles.includes(manageUserRolesName);
+    },
   },
   methods: {
-    roleArrayPosition: function(col, item) {
-      return (col - 1) * (this.itemsInColumn) + item - 1;
+    addRoles() {
```
tschia, line 254: We've been using the `:function () { .. }` format in most other places. For consistency we should do the same here.

frontend/src/components/UserUpdateRoles.vue
```diff
@@ -1,210 +1,384 @@
...
+      this.selectedClient = client;
+      this.isEdit = true;
+      this.dialog = true;
+      this.getUserClientRoles();
     },
-    getClients: function() {
-      return ClientsRepository.get()
-        .then(response => {
-          this.clients = response.data;
+    close() {
```
tschia, line 265: We've been using the `:function () { .. }` format in most other places. For consistency we should do the same here.

frontend/src/components/UserUpdateRoles.vue
```diff
@@ -1,210 +1,384 @@
...
+              return UsersRepository.getUserEffectiveClientRoles(
+                this.userId,
+                client.id
+              ).then((clientRoles) => {
+                clientRoles["clientRepresentation"] = client;
+                return clientRoles;
+              });
+            })
+          )
+            .then(function (rolesArray) {
```
tschia, line 299: I know this is from the copy/paste of the original method, and not something you added, but lets update this to use arrow notation like the rest of our methods
e.g. `(rolesArray) => {`

frontend/src/components/UserUpdateRoles.vue
```diff
@@ -1,210 +1,384 @@
...
+                    lastLogin: lastLoginStr,
+                  });
+                } else {
+                  clientsNoRolesAssigned.push(clientRoles.clientRepresentation);
+                }
+              });
+            })
+            .then((this.ClientsWithEffectiveRoles = results))
+            .then((this.clientWithoutRoles = clientsNoRolesAssigned))
+            .finally(() =>this.rolesLoaded = true);
```
tschia, line 324: These don't really need their own .then() blocks, they're not async methods
this could all be

`.finally(() => {
  this.ClientsWithEffectiveRoles = results
  this.clientWithoutRoles = clientsNoRolesAssigned
  this.rolesLoaded = true
})`

frontend/src/components/UserUpdateRoles.vue
```diff
@@ -1,210 +1,384 @@
...
+                    >
+                  </div>
+                </div>
+              </div>
+            </v-card>
+          </v-dialog>
+        </v-toolbar>
+      </template>
+      <template #item.lastLogin="{ item }">
+        <span v-if="item.lastLogin == -1">N/A</span>
```
sharpedavid, line 180: Avoid the use of [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants). (The anti-pattern refers to "numbers", but really any magic [literal](https://en.wikipedia.org/wiki/Literal_(computer_programming)) is suspect.

Options to consider:

- `item.lastLogin == 'NONE_ON_RECORD'` (string literal)
- `item.lastLogin == NONE_ON_RECORD` (constant variable)

Normally I'd favour using a constant variable, but "magic strings" can at least describe themselves (as opposed to `-1`).

frontend/src/components/UserUpdateRoles.vue
```diff
@@ -1,210 +1,384 @@
...
-      return ClientsRepository.get()
-        .then(response => {
-          this.clients = response.data;
+    close() {
+      this.dialog = false;
+    },
+    roleArrayPosition: function (col, item) {
+      return (col - 1) * this.itemsInColumn + item - 1;
+    },
+    addClientRoles: function () {
```
sharpedavid, line 271: My IDE reports `addClientRoles` as an unused property. Many IDE have this sort of scan built-in, but if one is unavailable in may be available via a plug-in or CLI tooling.

frontend/src/components/UserUpdateRoles.vue
```diff
@@ -1,210 +1,384 @@
...
+                v-show="
+                  selectedClient && (!clientRoles || clientRoles.length == 0)
+                "
+                type="article, button"
+              ></v-skeleton-loader>
+
+              <!-- role list -->
+              <div v-if="selectedClient">
+                <div
+                  width="inherit"
```
sharpedavid, line 57: My IDE reports that `width` is not a valid `div` attribute. I think it's probably right, reading [StackOverflow](https://stackoverflow.com/questions/2185519/what-are-valid-attributes-for-the-div-element-in-html).

`Attribute width is not allowed here`

frontend/src/components/UserUpdateRoles.vue
```diff
@@ -1,210 +1,384 @@
...
+                    v-model="selectedClient"
+                    @change="getUserClientRoles()"
+                    :disabled="isEdit"
+                  ></v-autocomplete>
+                </v-col>
+              </v-row>
+              <v-skeleton-loader
+                ref="roleSkeleton"
+                v-show="
+                  selectedClient && (!clientRoles || clientRoles.length == 0)
```
sharpedavid, line 49: Use `===` to avoid accidental type conversion.

frontend/src/components/UserMailboxAuthorizations.vue
```diff
@@ -0,0 +1,95 @@
...
+  data() {
+    return {
+      mailboxClientNames: ["SFDS", "HSCIS"],
+      clientsWithRoles: [],
+      selectedClient: 0,
+    };
+  },
+  async created() {
+    this.$root.$refs.UserMailboxAuthorizations = this;
+    this.getMailboxClients();
```
sharpedavid, line 41: My IDE reports "Missing await for an async function call."

sharpedavid (changes requested):
I code reviewed and tested this. The implementation seems logical and it is behaving well. I found a couple small edge case bugs. Please see comments.

frontend/src/components/UserUpdateRoles.vue
vangeliq: agree with the naming, but placing it on <style scoped> will make them ignore this style

sharpedavid (approved):

ahoplock:
The FMDB and WEBCAPS clients are spilling over for me locally with vertical scrollbars in both Firefox and Chrome but for only those clients.
image
image
...scrolling over just scrolls the dialog to the left?

frontend/src/components/UserUpdateRoles.vue
```diff
@@ -218,34 +389,54 @@ export default {
...
+  },
 };
 </script>
 
 <style>
-.right-gutters .col {
-  padding: 10px 12px 10px 0;
+.updateRolesDialog {
+    margin: 24px;
+    /* overflow-y: auto;
```
ahoplock, line 421: These commented out CSS properties are why the FMDB and WEBCAPS clients are spilling over as per my review comment. I'll make the change directly.

ahoplock:
I think this is fine to merge as it's been reviewed by David, Trevor, and myself at this point. Any further work (namely the SFDS API's being called every User Role update) can be handled in a separate bug/story.

PR #144, opened 2022-06-16
username whitespace gets trimmed

vangeliq:


vangeliq:
the whitespace on the start and end gets trimmed, space in the middle does not

ahoplock (approved):
Reviewed and approved. Tested locally; will deploy to Dev.

PR #145, opened 2022-06-23
session only times out when refresh token expires

vangeliq:


sharpedavid (approved):
Approved with minor revisions.

I am currently testing the changes.

sharpedavid:
I tested it and it seems to be working.

PR #153, opened 2022-08-16
Bug fixes 

vangeliq:
Fixed bugs 7201 and 7246


ahoplock:
The fix for 7246 doesn't appear to be working for me locally. I see no difference in behaviour between my local and UMC Dev. Which user are you testing this on? I was testing with `ahoplo_a@idir` and I see an empty org regardless of the environment.

ahoplock (approved):
Reviewed and approved. Will build and deploy to Dev.

PR #161, opened 2022-08-30
replaced user counts with a pie chart

vangeliq:
legend and piechart are not aligned to the left


frontend/src/components/PieChart.vue
```diff
@@ -0,0 +1,112 @@
...
+          data: this.pieChartData["UNIQUE_USER_COUNT"],       
+          datalabels: {
+            align: "start",
+            offset: -17,
+            anchor: function(context) {
+              let sum = 0;
+              for(let x in context.dataset.data){
+                sum = sum + context.dataset.data[x];
+              }
+              var index = context.dataIndex;
```
filipflorek, line 49: In general, try to avoid using `var`. They give trouble with scope sometimes. (For example you could access var declared in for loop outside of it, because it's scoped to the function)  So it's better to use `const `as a default, or `let` when you know that the variable is going to change.

filipflorek (commented):
Two minor comments, rather from the clean code perspective than functionality.

filipflorek (approved):

ahoplock (approved):
Reviewed and approved. Will merge and eventually deploy to Dev/Test (currently blocked by PHSA build).

PR #172, opened 2022-09-19
Ab#6892 users permissions tooltip

filipflorek:
Didn't know how to structure folders - I put model, mapper, service in one directory, should I split it for three separate ones?


backend/src/main/java/ca/bc/gov/hlth/mohums/model/GroupService.java
```diff
@@ -0,0 +1,74 @@
...
+        List<String> roleKeyWords = userManagementRoles.stream().map(role -> {
+                    String[] splitString = role.split("-", 3);
+                    return splitString[splitString.length - 1];
+                })
+                .distinct()
+                .collect(Collectors.toList());
+
+        String rolesString = String.join(", ", roleKeyWords);
+        return prefix + rolesString + ".\n";
+    }
```
sharpedavid, line 73: I feel like this class is perfect for some simple unit tests. It has no outside dependencies: just inputs and outputs. That said, they're most helpful to write during development rather than after, so it's up to you whether to add them or not.

You should also considering adding one integration test to MoHUmsIntegrationTests.java. The framework there is pretty good, and unlike unit tests I think they're still useful to add even after the code is written.

backend/src/main/java/ca/bc/gov/hlth/mohums/webclient/WebClientService.java
```diff
@@ -46,7 +56,17 @@ public ResponseEntity<List<Object>> getUsersInRole(String clientId, String roleN
...
+        ArrayList<LinkedHashMap> groups = (ArrayList<LinkedHashMap>) response.getBody();
+        assert groups != null;
+        List<Group> groupList = groups.stream().map(g -> getGroupById(g.get("id").toString())).collect(Collectors.toList());
+        return ResponseEntity.of(Optional.of(groupList));
+    }
+    // Group details
+    private Group getGroupById(String groupId) {
+        String path = String.format("%s/%s", groupsPath, groupId);
+        ResponseEntity<Object> response =  get(path, null);
+        return groupMapper.mapToGroup((LinkedHashMap)response.getBody());
```
sharpedavid, line 69: `groupMapper` is used only here in this method, `getGroupById`. `GroupMapper` has no external dependencies, and it is not an expensive object to create. We don't need to mock it to test `WebClientService`.

For these reasons, I recommend _not_ using injection to add `groupMapper` to `WebClientService`, and I don't even recommend making it a class field at all. Instead, either instantiate `groupMapper` right here in the method, or even turn `mapToGroup` into a `static` method so the object doesn't need to be instantiated at all.

sharpedavid (changes requested):
The is looking great. I have a bunch of refactoring recommendations. Please let me know what you think; I'm happy to discuss anything. While I feel strongly about some of my recommendations, they are sometimes subjective. We can also include Adam as a sort of "tie breaker".

I like the use of the Stream API to simplify and shorten the code.

backend/pom.xml
```diff
@@ -45,6 +45,12 @@
             <artifactId>jasypt-spring-boot-starter</artifactId>
             <version>3.0.3</version>
         </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>1.18.24</version>
+            <scope>provided</scope>
+        </dependency>
```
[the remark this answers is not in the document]
filipflorek: You're right, I'll change it so we stay consentient. The main reason I used it for, was that it provides @Builder, and @Value - which is basically same as Record, that I couldn't use because of Java version. But as I'm using those literally once it makes sense to get rid of them.

ahoplock (approved):
Reviewed and approved; will merge and deploy to Dev/Test.

PR #179, opened 2022-09-29
connect umc with org-api

filipflorek:
Connect UMS to org-api(aws) and enable admins to insert end edit organizations.

application.yaml changes:
When deploying to dev/test/prod environment the `organizations-api-url` and `proxy.host` needs to be updated accordingly.
For local tests and development `proxy.type` should be set to DIRECT


sharpedavid (approved):

ahoplock (approved):
Reviewed and approved.

PR #180, opened 2022-10-03
Ab#7823 web client service refactor

filipflorek:


backend/src/main/java/ca/bc/gov/hlth/mohums/controller/ClientsController.java
```diff
@@ -1,34 +1,31 @@
 package ca.bc.gov.hlth.mohums.controller;
 
 import ca.bc.gov.hlth.mohums.util.AuthorizedClientsParser;
-import ca.bc.gov.hlth.mohums.webclient.WebClientService;
+import ca.bc.gov.hlth.mohums.webclient.KeycloakApiService;
+import org.springframework.http.ResponseEntity;
+import org.springframework.util.LinkedMultiValueMap;
+import org.springframework.util.MultiValueMap;
+import org.springframework.web.bind.annotation.*;
```
nathanielcx, line 8: Is there a reason you changed the import to this from importing the explicit annotations you needed?

backend/src/main/java/ca/bc/gov/hlth/mohums/webclient/KeycloakApiService.java
```diff
@@ -0,0 +1,134 @@
...
+    public ResponseEntity<List<Object>> getUsersInRole(String clientId, String roleName,
+                                                       MultiValueMap<String, String> queryParams) {
+        String path = String.format("%s/%s/roles/%s/users", clientsPath, clientId, roleName);
+        return externalApiCaller.getList(path, queryParams);
+    }
+
+    // Groups
+    public ResponseEntity<Object> getGroups() {
+        ResponseEntity<Object> response = externalApiCaller.get(groupsPath, null);
+        ArrayList<LinkedHashMap> groups = (ArrayList<LinkedHashMap>) response.getBody();
```
nathanielcx, line 55: Would it be worth having some error handling here incase that get happens to fail?

backend/src/main/java/ca/bc/gov/hlth/mohums/webclient/KeycloakApiService.java
```diff
@@ -0,0 +1,134 @@
...
+        List<Group> groupList = groups.stream().map(g -> getGroupById(g.get("id").toString())).collect(Collectors.toList());
```
nathanielcx, line 56: Is some error handling required here in case the get doesn't work properly? (this may not be the case so ignore this if a non-200 response is handled properly)

backend/src/main/java/ca/bc/gov/hlth/mohums/webclient/ExternalApiCaller.java
```diff
@@ -0,0 +1,89 @@
+package ca.bc.gov.hlth.mohums.webclient;
+
+import org.springframework.http.HttpMethod;
+import org.springframework.http.ResponseEntity;
+import org.springframework.util.MultiValueMap;
+import org.springframework.web.reactive.function.client.WebClient;
+
+import java.util.List;
+
+public class ExternalApiCaller {
```
nathanielcx, line 10: Should there be some unit testing for the methods in this class and the others that have been added?

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/ClientsController.java
filipflorek: Didn't do that on purpose - just realized that the keyboard shortcut I use for deleting unused imports, is 'optimizing' the imports as well - and grouped those 5 explicit into  `import org.springframework.web.bind.annotation.*;`
Will roll back to explicit ones.

backend/src/main/java/ca/bc/gov/hlth/mohums/webclient/KeycloakApiService.java
filipflorek: Was asking the same thing in the beginning, but the goal is to have it working as basic proxy, so we just handle error response from Keycloak API in frontend

backend/src/main/java/ca/bc/gov/hlth/mohums/webclient/ExternalApiCaller.java
sharpedavid: My thoughts in no particular order:

* My advice for this is to write unit tests when you find them useful for developing the functionality, and otherwise focus on integration or acceptance tests; that is, calling the actual REST API.

* A test in `MoHUmsIntegrationTests` will cover `ExternalApiCaller`.

* The advantage to focusing on acceptance tests is that they only to be changed if the business requirements change, for example if the contract of the REST API changes. Test cases should not be tightly coupled to the structure of the code.

* If someone were to refactor the application in the future, they might just throw out ExternalApiCallerTest because the structure would change. They wouldn't throw out `MoHUmsIntegrationTests`.

* I do write unit tests, but primarily as a "test driven development" activity. I write them before I write the code as a sort of outline for what I'm trying to do. Sometimes I delete them or change them into acceptance tests if it's appropriate.

Anyhow, I hope some of that is useful. I'm not asking for any changes.

backend/pom.xml
```diff
@@ -45,6 +45,18 @@
...
+            <artifactId>okhttp</artifactId>
+            <version>4.0.1</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
+            <artifactId>mockwebserver</artifactId>
+            <version>4.0.1</version>
+            <scope>test</scope>
+        </dependency>
```
sharpedavid, line 59: Please move these down with the rest of the `test` dependencies.

backend/pom.xml
```diff
@@ -45,6 +45,18 @@
             <artifactId>jasypt-spring-boot-starter</artifactId>
             <version>3.0.3</version>
         </dependency>
+        <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
+            <artifactId>okhttp</artifactId>
+            <version>4.0.1</version>
+            <scope>test</scope>
+        </dependency>
```
sharpedavid, line 53: For `test` dependencies I'm less particular, but in some cases I would request using what's already available if possible. In this case that would be [`org.springframework.boot.test.web.client.TestRestTemplate`](https://www.baeldung.com/spring-boot-testresttemplate). Here I'm mentioning it just so you're aware.

If we use the "standard" libraries, be they Spring or Core Java:

* It compiles a touch faster.
* File size is a bit smaller.
* People are likely to be familiar with the libraries you're using.
* Fewer dependencies is arguably less complex.
* Long term support is more likely.

I know these bullet points are one-sided; there are many reasons to reach for the best library for the job.

I'm not asking for changes here. Just sharing my thoughts.

sharpedavid (changes requested):
I approve these changes with one cosmetic request to rearrange the pom dependencies.

sharpedavid (approved):

frontend/src/components/UserSearch.vue
```diff
@@ -206,10 +206,15 @@
...
+                      <template v-slot:label>
+                        <v-tooltip v-if="clientRoles[roleArrayPosition(col, item)].description" right max-width="300px">
+                          <template v-slot:activator="{ on }">
+                            <span v-on="on">{{clientRoles[roleArrayPosition(col, item)].name}}</span>
+                          </template>
+                          <span class="white-space-fix">{{clientRoles[roleArrayPosition(col, item)].description}}</span>
+                        </v-tooltip> 
+                        <span v-else>{{clientRoles[roleArrayPosition(col, item)].name}}</span>        
+                      </template>
                   </v-checkbox>
```
sharpedavid, line 218: Were these changes intentional? They weren't in the original PR.

PR #244, opened 2023-03-09
AM-8691-create-umc-chart-login-over-time

alex-tremblay-cgi:


frontend/src/components/Dashboard.vue
```diff
@@ -161,50 +208,71 @@ export default {
...
+      const response = await MetricsRepository.get(
+        "total-active-user-count/" + format
+      );
+      const labels = [];
+      const dataset = [];
+      for (var key of response.data.entries()) {
+        labels.push(new Date(key[1]["DATEE"]).toISOString().split("T")[0]);
+        dataset.push(key[1]["ACTIVE_USER_COUNT"]);
+      }
+      this.totalUserCount["DATEE"] = labels;
```
filipflorek, line 246: As per my comment from `MetricController.java`  the DATEE should be more descriptive

frontend/src/components/Dashboard.vue
filipflorek, line 1: The charts look awesome, one thing I would look into, maybe brainstorm with other devs is the amount of API/ database calls we're performing here.
Whenever user selects a different chart scale (7 days, 1 month etc.) there's a new API call, right?
Maybe it's worth to save results to application memory on frontend, to limit the numbers of api calls? Just an idea, I haven't looked into how much of data is being sent, especially in production environment.

filipflorek (changes requested):
The charts look really nice, good work. Couple of minor suggestions, and one topic to discuss with other devs maybe. I still need to look into SQL queries, but that's what I have for now!

frontend/src/components/Dashboard.vue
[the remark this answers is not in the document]
alex-tremblay-cgi: Yes I used prettier for the files I change and my vscode autoformat on save, I will change that for now thank

alex-tremblay-cgi: I added a version that work using localStorage and refresh each day in MetricsRepository.js. If I could get a review on that. I was thinking of maybe adding a button to clear the cash also

filipflorek (changes requested):

frontend/src/components/Dashboard.vue
sharpedavid: @filipflorek's reasoning sounds right to me.

Also, not many people use the UMC. It is used by access administrators, so probably a few dozen people. Even fewer people have access to the Dashboard, and they use it infrequently.

So we don't need to worry about taxing our database with too many calls. Filip's idea of storing the data in component variables is the right one though. It will make the interface snappy and more usable.

frontend/src/components/Dashboard.vue
```diff
@@ -45,33 +48,37 @@
...
             </v-tooltip>
           </div>
           <p class="single-stat">{{ totalNumberOfUsers }}</p>
           <v-skeleton-loader
             v-if="totalNumberOfUsersLoadingStatus"
             ref="dashboardSkeleton"
             type="text"
-            max-width="60px">
+            max-width="60px"
+          >
```
sharpedavid, line 63: I assume this accidental, but just a reminder to avoid including formatting changes in PR. If you want to do formatting, open a second PR later.

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/MetricsController.java
```diff
@@ -58,6 +59,36 @@ public List<Map<String, Object>> getActiveUserCount() throws SQLException {
...
+    @GetMapping("/metrics/total-active-user-count/{format}")
+    public List<Map<String, Object>> getTotalActiveUserCountWeek(@PathVariable String format) throws SQLException {
+        String sqlBase
+                = " SELECT EVENT_DATE, COUNT(1) AS ACTIVE_USER_COUNT"
+                + " FROM ("
+                + "    SELECT TRUNC(to_date('19700101', 'YYYYMMDD') + ( 1 / 24 / 60 / 60 / 1000) * event_time) AS EVENT_DATE"
+                + "    FROM keycloak.event_entity"
+                + "    WHERE type = 'LOGIN'"
+                + "  )"
+                + " %s"
```
sharpedavid, line 71: The way you have implemented this is safe, but please be wary of inserting user-provided data into SQL strings. I realize you are not inserting the user-provided `path` in this code, so the advice is not immediately applicable. 

https://owasp.org/www-community/attacks/SQL_Injection
https://xkcd.com/327/

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/MetricsController.java
```diff
@@ -58,6 +59,36 @@ public List<Map<String, Object>> getActiveUserCount() throws SQLException {
         return jdbcTemplate.queryForList(sql);
     }
 
+    @GetMapping("/metrics/total-active-user-count/{format}")
+    public List<Map<String, Object>> getTotalActiveUserCountWeek(@PathVariable String format) throws SQLException {
```
sharpedavid, line 63: We have integration tests at `moh-keycloak-user-management\backend\src\test\java\ca\bc\gov\hlth\mohums\integration\MoHUmsIntegrationTests.java`. Please add at least a smoke test for this SQL query.

As it stands, a simple typo in the query would go unnoticed until someone manually tests it. A "smoke test" is a low cost way to identity trivial but irritating and potentially costly mistakes.

https://en.wikipedia.org/wiki/Smoke_testing_(software)

sharpedavid (dismissed):
A great contribution. I'm happy to see LineChart in its own component, and the changes you've made are concise and easy to understand.

Please avoid including formatting changes in the future.

All of my feedback is for discussion and consideration.

frontend/src/components/Dashboard.vue
```diff
@@ -82,26 +89,57 @@
...
+            type="image"
+          >
           </v-skeleton-loader>
         </div>
       </div>
     </div>
+    <div class="flex">
+      <div class="tile" style="width: 1100px">
+        <div class="heading">
+          <p>Active Total User Count</p>
```
sharpedavid, line 105: I think "Active Users" is ambiguous so I'm trying to think of a better name. How about "User Login Events per Day"?

Also, could you experiment a little with reducing the number of X axis labels?

E.g. Before:

image

After:
image

https://stackoverflow.com/a/68139050/201891

frontend/src/components/Dashboard.vue
alex-tremblay-cgi: Perfect I will switch the implementation to a variable, thanks for the input!

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/MetricsController.java
```diff
@@ -58,6 +59,36 @@ public List<Map<String, Object>> getActiveUserCount() throws SQLException {
...
+                return jdbcTemplate.queryForList(String.format(sqlBase, sqlWhereMonth));
+            case "6M":
+                return jdbcTemplate.queryForList(String.format(sqlBase, sqlWhereSixMonths));
+            case "1Y":
+                return jdbcTemplate.queryForList(String.format(sqlBase, sqlWhereYear));
+            default:
+                return jdbcTemplate.queryForList(String.format(sqlBase, sqlWhereWeek));
+        }
+    }
+
```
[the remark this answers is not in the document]
alex-tremblay-cgi: (1) I spoke with Filip yesterday about this. The request to get a year of data take around 3s and it's less than a second for a week. 
(2) It's always fetch each time we go to the dashboard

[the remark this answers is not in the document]
alex-tremblay-cgi: Filip tested with the PROD database and here are the time they take to process (only for database need to add time for the backend)
7 days -> 1.2s
1 month -> 1.5s
6 months -> 3.6s
12 months -> 3.9s

Right now, for anyone who go on the dashboard the 7 days will be loaded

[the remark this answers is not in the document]
sharpedavid: I think 4 seconds is acceptable. 

The skeleton loader means the page is loaded right away while the data is populated, and the chart will be used infrequently, so 4 seconds isn't too long to wait. After the 4 second load, users could switch between each choice (7 days etc.) without additional fetches or database queries.

All this said, it's fine the way it is, but it does increase the complexity a bit without significant benefit. If it's easy to change and filter with the data on the frontend, I'd recommend it.

frontend/src/api/MetricsRepository.js
```diff
@@ -1,9 +1,51 @@
...
 
 export default {
     get(queryParams) {
         return umsRequest().then(axiosInstance => axiosInstance.get(`${resource}/${queryParams}`));
-    }
-}
+    },
+    async getTotalActiveUserCount(format) {
+        let yearOfActiveUserCount = Vue.prototype.UserCountCache;
+        if (Object.keys(yearOfActiveUserCount).length === 0) {
```
filipflorek, line 12: `Object.keys` method throws error when `yearOfActiveUserCount ` is undefined - charts won't load.
I changed it locally to `if (yearOfActiveUserCount === undefined || Object.keys(yearOfActiveUserCount).length === 0)` to make the charts load, but there's probably cleaner way to do it. (I didn't dig into the length check)

filipflorek (changes requested):
The charts are looking neat, and they load up fast, data is kept after changing between tabs. Great work! There's one bug, but easy to fix, that prevent the charts from loading.

frontend/src/api/MetricsRepository.js
alex-tremblay-cgi: I think It's because I forgot to commit `Vue.prototype.UserCountCache = {};` in main.js ... gonna do that right now and check if the bug is still present

filipflorek (approved):
Nice work, thanks for making the changes

PR #251, opened 2023-04-14
Setup prettier

filipflorek:


alex-tremblay-cgi:
I think we need the file `.prettierrc` because else it will default to vscode settings and if multiple people have change their setting the action will not pass

frontend/package.json
```diff
@@ -8,7 +8,9 @@
     "test:unit": "vue-cli-service test:unit",
     "test:e2e": "testcafe chrome tests/e2e",
     "lint": "vue-cli-service lint",
-    "test": "npm run test:unit"
+    "test": "npm run test:unit",
+    "format:check": "prettier -c --loglevel debug --config-precedence prefer-file ./src",
```
alex-tremblay-cgi, line 12: Maybe we can remove loglevel now that our problem is fixed, it will generate a more clean input of non formatted files

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,26 @@
...
+          ref: ${{ github.head_ref }}
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 14.16.1
+
+      - name: Check code format
+        run: |
+          npm ci
```
alex-tremblay-cgi, line 25: Here maybe it would be more clean to have a separate action for `npm ci` vs `npm run format:check` so that in the github action tab we have a clean output of file that need formatting (and that doesn't start with all the logs from `npm ci`)

alex-tremblay-cgi (commented):
Nice job on making prettier work with GitHub action!

alex-tremblay-cgi (approved):
Nice job on making it work!

PR #256, opened 2023-05-18
Add ability to assign MSP Payee number

weskubo-cgi:
See https://dev.azure.com/cgi-vic-hlth/AM%20Team/_workitems/edit/7203.

UMC
- Added new component and integrated with existing User Roles popup
- Added Service method
- Added configuration

UMS
- Added new endpoint. For simplicity a single PUT endpoint was defined for adding/updating/removing the value. UMS will determine which MSP Direct endpoint to call.
- Added code for calling MSP Direct.
- Added integration test for MSP Direct Payee Mapping API

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,5 @@ target/
 /frontend/nbproject/private/
 frontend/dist/*
 frontend/nbproject/*
+
+yarn.lock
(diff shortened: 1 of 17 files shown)


filipflorek (approved):
Looks good, thanks for writing that feature.

PR #262, opened 2023-08-01
update IDP name mappings

filipflorek:
See https://dev.azure.com/cgi-vic-hlth/AM%20Team/_sprints/taskboard/AM%20Team/AM%20Team/Sprint%2089?workitem=10358 for details.

The mapping is hardcoded, following the previous solution. It would be possible to utilize the Keycloak API to return `Display Name` of the federated Identity Provider.


ahoplock (approved):
Reviewed and approved.

PR #264, opened 2023-08-29
Bcmoham 13875 last logon date not working in keycloak report

daveb-hni:
Updated the Advanced Search to use the client id instead of client name as search criteria because that is what is stored in the tables that it searches against. i.e. event_entity and client.

### DIFF
--- a/backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
+++ b/backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
@@ -72,21 +72,21 @@ public class UsersController {
 
     /**
      * 
-     * @param briefRepresentation
-     * @param email
-     * @param first
-     * @param firstName
-     * @param lastName
-     * @param max
-     * @param search
-     * @param username
-     * @param org
-     * @param lastLogAfter
-     * @param lastLogBefore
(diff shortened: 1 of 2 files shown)


backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
```diff
@@ -130,17 +148,17 @@ public ResponseEntity<List<Object>> getUsers(
             }
             Optional<String> clientAndRolesJoins = Optional.empty();
             Optional<String> clientAndRolesCriteria = Optional.empty();
-            if (clientName.isPresent() && clientId.isPresent() && clientNameAndIdAreValid(clientName.get(), clientId.get())) {
+            if (clientClientId.isPresent() && clientId.isPresent() && clientIdAndIdAreValid(clientClientId.get(), clientId.get())) {
```
weskubo-cgi, line 151: In the future it would make sense to only pass a single identifier for the client (e.g. id) in the API call and then lookup the other values (e.g. clientId) via an API/db call. It also shouldn't be necessary to validate the values "match" in that case.

weskubo-cgi (changes requested):

weskubo-cgi (approved):

PR #273, opened 2024-05-10
Umc UI updates

filipflorek:
### Changes being made

Renamed Create User buttons. Added success alert in Mailbox Authorization component. Refactored code that renders username tooltip.

### Context

PR against Advanced search branch, as that's the one currently in dev and test.
Renaming only button names, to make it more clear for end users. Refactoring would have to include component name changes and router updates, without much added value (it technically is "creating user" from dev perspective) - Please let me know if leaving URL as `user/create` is acceptable.

### DIFF
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -6,6 +6,8 @@ The "User Management Console" is custom user management console for Keycloak. Ke
 
 # Prerequisites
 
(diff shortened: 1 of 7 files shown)


ahoplock (approved):
Reviewed and approved.

PR #276, opened 2024-05-24
add alias mapping for PHO-RSC clients

filipflorek:
### Changes being made

Added a function in ClientRepository.js that swaps clientId to defined alias, for PHO-RSC and PHO-RSC-GROUPS clients.

### Context

This is a UI change only, following the Advanced Search upgrade, all the API calls are using ID instead of clientID.
PHO-RSC client is rebranding, referenced as Posit from now on. The client request was to map the name in dropdowns from:
PHO-RSC ->POSIT-USER-ROLES
PHO-RSC -> POSIT-GROUP-ROLES

### Quality Check

- [x] E2E/ Unit/ Integration tests have been added.
- [x] Frontend tests have been successfully run.
- [x] Backend tests have been successfully run.

### DIFF
--- a/frontend/src/api/ClientsRepository.js
+++ b/frontend/src/api/ClientsRepository.js
@@ -22,4 +22,20 @@ export default {
       )
     );
   },
+
+  assignClientAliases(clients) {
+    let clientAliases = [
+      { clientId: "PHO-RSC", alias: "POSIT-USER-ROLES" },
+      { clientId: "PHO-RSC-GROUPS", alias: "POSIT-GROUP-ROLES" },
+    ];
+    return clients.map((client) => {
+      const aliasMapping = clientAliases.find(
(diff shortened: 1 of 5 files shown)


frontend/src/api/ClientsRepository.js
```diff
@@ -22,4 +22,20 @@ export default {
       )
     );
   },
+
+  assignClientAliases(clients) {
+    let clientAliases = [
```
ahoplock, line 27: Looks like this is just using an array as a map, so it could probably just be that instead which would simplify things and might be more performant (map `.get()` vs. array `.find()`). That being said, it's a small array and likely not going to grow much past this, so this is more of an "FYI" comment instead of a "do this or I will not approve" comment.

ahoplock (dismissed):
Reviewed and approved.

ahoplock (approved):
Reviewed and approved.

PR #277, opened 2024-06-05
Umc last login and group description fix

filipflorek:
### Changes being made

1. Fix group description mapping in UMC.
2. Added User Last Login for client aliases, so that the date displays in "User Details" page.
3. Fix last login date display (https://dev.azure.com/cgi-vic-hlth/AM%20Team/_sprints/taskboard/AM%20Team/AM%20Team/Sprint%20105?workitem=11114)

### Context

1. Issue with group description mapping was caused by how String.replace() is executed. Solution was to create regex. 
  Could not use `\\b` character (word boundary) as it includes only letters, numbers and underscore. The client aliases can include also `-` that's why regex is checking for this `(?!\\w|-)`
  So description would look like this:
  image
  Instead of this:
  image

2. User last login was not displaying in user details page for clients with aliases (POSIT), because alias was not mapped to the LastLogin object.

3. Last login was not showing up because UMS returns <client id, timestamp> and UMC expected <client name, timestamp>

### Quality Check

- [x] Frontend tests have been successfully run.
- [N/A] Backend tests have been successfully run.
 - [N/A] E2E/ Unit/ Integration tests have been added.

### DIFF
--- a/frontend/src/api/ClientsRepository.js
+++ b/frontend/src/api/ClientsRepository.js
@@ -23,6 +23,11 @@ export default {
     );
   },
 
+  /**
+   * Some clients request that their application is referred by different name than Keycloak's Client Id
+   * Because changing Client Id requires updating authentication flow on the client side, UMC client aliases are intorduced.
+   * Define entry in the below array to display alias instead of clientID for a given client.
+   */
   clientAliases: [
     { clientId: "PHO-RSC", alias: "POSIT-USER-ROLES" },
     { clientId: "PHO-RSC-GROUPS", alias: "POSIT-GROUP-ROLES" },
--- a/frontend/src/api/ClientsRepository.js
+++ b/frontend/src/api/ClientsRepository.js
@@ -24,8 +24,8 @@ export default {
   },
 
   /**
-   * Some clients request that their application is referred by different name than Keycloak's Client Id
-   * Because changing Client Id requires updating authentication flow on the client side, UMC client aliases are intorduced.
+   * Some clients request that their application is referred to by a different name than Keycloak's Client Id
+   * Because changing Client Id requires updating authentication flow on the client side, UMC client aliases are introduced.
    * Define entry in the below array to display alias instead of clientID for a given client.
    */
   clientAliases: [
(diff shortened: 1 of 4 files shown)


filipflorek:
So far in UMC we have only E2E tests, is it worth configuring unit test suite? For cases like regex?

frontend/src/components/UserUpdateRoles.vue
```diff
@@ -344,11 +346,10 @@
...
               rolesArray.forEach((clientRoles) => {
                 if (clientRoles.data.length > 0) {
-                  let lastLoginStr = LAST_LOGIN_NOT_RECORDED;
-                  if (lastLoginMap[clientRoles.clientRepresentation.name]) {
-                    lastLoginStr =
-                      lastLoginMap[clientRoles.clientRepresentation.name];
-                  }
+                  let lastLoginStr =
+                    lastLoginMap[clientRoles.clientRepresentation.clientId] ||
+                    LAST_LOGIN_NOT_RECORDED;
```
sharpedavid, line 351: You don't need to change it, but just for future consideration I do recommend explicitly checking `!== undefined` instead of relying on "truthy" values (undefined, null, empty string).

sharpedavid (changes requested):
Documentation request.

I tested it though, and it's working. I tested with POSIT, EMCOD, and a couple of other searches.

frontend/src/api/ClientsRepository.js
```diff
@@ -23,6 +23,11 @@ export default {
     );
   },
 
+  /**
+   * Some clients request that their application is referred by different name than Keycloak's Client Id
+   * Because changing Client Id requires updating authentication flow on the client side, UMC client aliases are intorduced.
```
sharpedavid, line 28:
```suggestion
   * Because changing Client Id requires updating authentication flow on the client side, UMC client aliases are introduced.
```

frontend/src/api/ClientsRepository.js
```diff
@@ -23,6 +23,11 @@ export default {
     );
   },
 
+  /**
+   * Some clients request that their application is referred by different name than Keycloak's Client Id
```
sharpedavid, line 27:
```suggestion
   * Some clients request that their application is referred to by a different name than Keycloak's Client Id
```

sharpedavid (approved):

sharpedavid (approved):

PR #293, opened 2024-10-28
Add group membership report

sharpedavid:
### Changes being made

Adds a "Group Membership" tab, which displays group details along with their members.

### Context

[BCMOHAD-25629](https://proactionca.ent.cgi.com/jira/browse/BCMOHAD-25629)

### Quality Check

- [ ] E2E/ Unit/ Integration tests have been added.
- [ ] Frontend tests have been successfully run.
- [x] Backend tests have been successfully run.


filipflorek (changes requested):
Thanks for making clear and concise changes. The new tab looks good. I requested some changes regarding the table footer and have some questions about permissions to the new tab.

Most administrators can only manage the groups that they belong to. So what they see on `/groupreport` page and user page is consistent. 
An edge case would be Access Management Team since we have "manage-all-groups" role On the user screen I can assign user to any of the existing group, but `/groupreport` shows only those that I belong to.  This is a note, since Access Management Team should not be handling group assignments anyway.

filipflorek (approved):

PR #300, opened 2025-04-16
Bcmohad 27789 update ums

msdkgit:
### Changes being made

Implement new "read-only-client" role policy, Allow admin to edit or add application role only if he has a "view-client-" role

### Quality Check

- [ ] E2E/ Unit/ Integration tests have been added.
- [ ] Frontend tests have been successfully run.
- [ ] Backend tests have been successfully run.
- [ ] The actual changes are separated from formatting changes.


backend/src/test/java/ca/bc/gov/hlth/mohums/controller/UsersControllerTest.java
```diff
@@ -55,5 +64,44 @@ public void testUserLocation_badGuid() {
         // If the Location header doesn't match what we expect, namely /users/VALID_GUID, don't touch it.
         assertEquals("https://common-logon-dev.hlth.gov.bc.ca/auth/admin/realms/moh_applications/users/NOT_A_GUID", newHeaders.getLocation().toASCIIString());
     }
+    
+     //PLR 	   : dc7b9502-3ffa-4ff8-be2e-ebfebe650590
+    // BCER-CP : 9ee2f550-307a-414a-bda4-d9c44235a2a4
```
sharpedavid, line 69: Use named constants for these values and then you won't need comments because it will be "self documenting".

backend/src/test/java/ca/bc/gov/hlth/mohums/controller/UsersControllerTest.java
```diff
@@ -2,15 +2,24 @@
...
 import org.springframework.http.HttpHeaders;
+import org.springframework.test.context.junit4.SpringRunner;
 
 import java.net.URI;
 
 import static org.junit.jupiter.api.Assertions.*;
 
+@RunWith(SpringRunner.class)
+@SpringBootTest
 class UsersControllerTest {
```
sharpedavid, line 17: Thanks for adding the tests around `isAuthorizedToEditClient` — they help validate the logic in isolation. That said, I think we should shift the focus to testing the behavior through actual API calls.

We already have patterns for this in `MoHUmsIntegrationTests` (e.g., `deleteUserClientRole`), and I’d prefer to see the new read-only behavior tested there. These integration tests are a bit more work to set up, but they test the *actual feature* and give us much higher confidence. They also make sure the method is actually used in the request path.

Once that coverage is in place, we can safely make `isAuthorizedToEditClient` private and delete the unit tests — the behavior will be fully covered externally. That keeps the test suite leaner and more maintainable in the long run.

Happy to help sketch out an integration test case if that’s useful.

backend/src/main/java/ca/bc/gov/hlth/mohums/util/AuthorizedClientsParser.java
```diff
@@ -20,6 +20,11 @@ public List<String> parse(String token) {
         JSONArray roles = parseRoles(token);
         return parseClients(roles);
     }
+    
+    public List<String> parseToEdit(String token) {
```
sharpedavid, line 24: I'd rename the other method. We have `parse` and `parseToEdit`, so `parse` should be named something like `parseToReadOnly` or something like that.

backend/src/main/java/ca/bc/gov/hlth/mohums/util/AuthorizedClientsParser.java
```diff
@@ -40,7 +45,26 @@ private JSONArray parseRoles(String token) {
...
+			} else {
+				parts = ((String) role).split("read-only-client-");
+				if (parts.length == 2) {
+					clients.add(parts[1].toLowerCase());
+				}
+			}
+		});
+
+		return clients;
+	}
```
sharpedavid, line 65: This might just be a personal preference, but I find the use of `split` and `parts.length` a bit obtuse here. It’s doing an implicit check on the role prefix based on how the string splits, when it could be clearer to just check the prefix directly.

Something like this feels easier to read and reason about:

```java
private List<String> parseClients(JSONArray roles) {
    List<String> clients = new ArrayList<>();

    roles.forEach(roleObj -> {
        String role = (String) roleObj;

        if (role.startsWith("view-client-")) {
            clients.add(role.substring("view-client-".length()).toLowerCase());
        } else if (role.startsWith("read-only-client-")) {
            clients.add(role.substring("read-only-client-".length()).toLowerCase());
        }
    });

    return clients;
}
```

It also avoids weird edge cases like this:

```java
"not-a-real-role-containing-view-client-hcimweb"
```

With the current `split` logic, this would get parsed and add `"hcimweb"` to the client list — even though it’s not a valid role. Probably not something we’ll run into often, but using `startsWith` avoids the risk entirely.

sharpedavid (changes requested):

backend/src/test/java/ca/bc/gov/hlth/mohums/controller/UsersControllerTest.java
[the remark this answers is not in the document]
msdkgit: fixed

backend/src/main/java/ca/bc/gov/hlth/mohums/controller/UsersController.java
```diff
@@ -373,6 +373,21 @@ boolean isAuthorizedToViewClient(String token, String clientGuid) {
         }
 
     }
+    
+    boolean isAuthorizedToEditClient(String token, String clientGuid) {
+        AuthorizedClientsParser acp = new AuthorizedClientsParser();
+        List<String> authorizedClients = acp.parseToEdit(token);
+
+        LinkedHashMap<String, String> client = (LinkedHashMap<String, String>) keycloakApiService.getClient(clientGuid).getBody();
+        // TODO If the client doesn't exist return a 401 (should this be a 404 which is the actual KC response)
```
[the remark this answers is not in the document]
msdkgit: fixed

backend/src/main/java/ca/bc/gov/hlth/mohums/util/AuthorizedClientsParser.java
msdkgit: fixed

backend/src/main/java/ca/bc/gov/hlth/mohums/util/AuthorizedClientsParser.java
msdkgit: fixed

sharpedavid (approved):

